PR #2, opened 2016-02-08
remove active_job, active_record, and action_mailer from required frameworks

jmartin-sul:
since this is supposed to be a slimmed down application focused on just indexing things from fedora to solr, and we want to be able to spin instances up easily to meet high indexing loads, it seems to make sense to not include things unnecessarily.

i imagine we won't be using the following in this app in the near future:
* active_job: i could see adding this back if we decide to have dor_indexing_app handle the bulk indexing that argo does now, though that's slightly trickier to extricate, and really what we want to do there is slim down the size of the workers that do the indexing, which doesn't necessarily mean moving it outside of argo.
* active_record: no need to do any database stuff at this point.  i get the impression that no session management is needed here, since authentication between dlss services seems to happen via IP whitelisting and/or certificate-based auth.
* action_mailer: no plans to send email.

there are notes as to what corresponding settings were commented out, since they would only be relevant to the things that were removed, and caused errors when trying to start the application without the frameworks to which they pertained.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -3,10 +3,13 @@ require File.expand_path('../boot', __FILE__)
 require "rails"
 # Pick the frameworks you want:
 require "active_model/railtie"
-require "active_job/railtie"
-require "active_record/railtie"
+#require "active_job/railtie"
+#NOTE: if active_record is added back, uncomment config.active_record.raise_in_transactional_callbacks
+# below, and ActiveRecord::Migration.maintain_test_schema! in spec/rails_helper.rb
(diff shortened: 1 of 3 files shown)


jmartin-sul:
open to not doing this if people have objections, or other ways of doing this if people have alternative suggestions.

cbeer:
:+1: makes sense to me. When we go to add capistrano to the application, we might have to do something to disable e.g. database migrations too.

mejackreed:
:+1:

PR #7, opened 2016-02-10
add capistrano gems and configuration

jmartin-sul:
* add capistrano gems
* add the config files that Capistrano uses to define deployments
 * Capfile and config/deploy.rb provide the base for all deployments
 * add stage as the first specific deployment

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,0 +1,30 @@
+# Load DSL and Setup Up Stages
+require 'capistrano/setup'
+
+# Includes default deployment tasks
+require 'capistrano/deploy'
+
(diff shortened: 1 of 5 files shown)


jmartin-sul:
just to see what would happen, i also tried doing the `bundle exec cap install` autogeneration (https://github.com/capistrano/capistrano#capify-your-project).  but when i diffed that test branch against this one, it didn't seem to contain anything i'd missed by assembling the files by hand based on existing dlss code as examples (argo and searchworks).

note that stage is the only deployment environment that's currently defined by this PR.  we can add dev and prod once this change has resulted in a successful deployment.

PR #9, opened 2016-02-18
set up stage deployment for load balancing, 

jmartin-sul:
also, 
* set deployment user explicitly at top level
* remove unneeded db role since there's no db for this app.

@eefahy @mejackreed


mejackreed:
Works by me. @eefahy cool to deploy them all at once?

jmartin-sul:
as per @eefahy, also set the user explicitly (just pushed a commit for that)

PR #13, opened 2016-02-20
switch to is_it_working-cbeer and add fedora and solr checks

jmartin-sul:
unfortunately, this is not working yet.  tests fail because it looks to load cert files, but can't, because there are no cert files in the default installation.  runs fine in server mode under the dev env on my laptop (which does have properly configured certs).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'squash_rails'
 gem 'squash_ruby'
 
 gem 'dor-services', '~> 5.0', '>= 5.3.4'
-gem 'is_it_working'
+gem 'is_it_working-cbeer'
 
 group :development, :test do
   gem 'rspec-rails', '~> 3.0'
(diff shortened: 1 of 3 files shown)


jmartin-sul:
should be working now.  an example of the fuller from `/is_it_working` after this change:
```
Host: dor-indexing-app-stage-b.stanford.edu
PID:  24046
Timestamp: 2016-02-22T17:52:25-08:00
Elapsed Time: 204ms

OK   : rubydora - service active (10ms)
INFO : rubydora - repositoryName - Fedora Repository (10ms)
INFO : rubydora - repositoryBaseURL - https://localhost:443/fedora (10ms)
INFO : rubydora - repositoryVersion - 3.8.1 (10ms)
OK   : rsolr - service active (195ms)
INFO : rsolr - numDocs: 207799 (195ms)
INFO : rsolr - lastModified: 2016-02-22T22:36:14.924Z (195ms)
```

if there are cleaner ways of doing this than only configuring `IsItWorking::Handler` for non-test environments, i'm open to changing the approach.

@eefahy @cbeer

PR #15, opened 2016-02-23
Add simplecov to get percentage coverage report on laptop runs

jmartin-sul:


jmartin-sul:
not actually showing percentage covered (in PR or on laptop).  odd.  will investigate...

drh-stanford:
is this PR still being worked on? if not, shall we delete the branch?

jmartin-sul:
i don't think it's being worked on actively, though it is still something i'd like to add.  i'd prefer to leave the branch and PR in limbo until we can come back around to it, since the branch list for this project isn't that cluttered, though i also wouldn't be too torn up if it were deleted.

cbeer:
I've rebased this and got it working.

PR #21, opened 2016-04-11
update to latest (5.8.0) dor-services

jmartin-sul:
* bundle update
* update config file to switch over from deprecated field name

closes sul-dlss/media#23


ndushay:
:+1:

PR #30, opened 2016-09-21
update dor-services and other dependencies

jmartin-sul:
update to the latest dor-services (5.10.2) so that we can get the version that uses the latest stanford-mods (2.2.2), so that indexing of searchworks genre field matches present searchworks indexing behavior.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem 'rails', '4.2.7.1'
 gem 'squash_rails'
 gem 'squash_ruby'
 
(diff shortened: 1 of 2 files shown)


ndushay:
since this upgrade masks a ginormous increase in stanford-mods release, including non-backwards compatible changes, I think that maybe 

- dor-services gem might have higher version bump????
- this PR should include any necessary changes to indexing based on stanford-mods changes.  

Have discussed the later point with @jmartin-sul  F2F ...

PR #42, opened 2016-10-20
Update to Rails 5

cbeer:


drh-stanford:
i've rebased the branch and the PR no longer has conflicts.

cbeer:
Deployed to -stage. Given how little data is moving through there, I'm not sure how much we're going to learn, though.

When deploying to -prod, if we're still worried, we could deploy to only one of the nodes for a little bit. I suspect that's overly cautious though. We've now updated a good number of applications to Rails 5, and with a team actively working on the project, if we should get exception notifications, we'll be able to roll it back and fix the problem in a timely manner.

PR #58, opened 2016-11-03
port monitoring from is_it_working to OkComputer

### LINKED ISSUE #48: port to OkComputer
see #44. need to port the existing `is_it_working` implementation to OkComputer using the following routes:

- `/status/default` -- a ping
- `/status/version` -- echo of the installed version number
- `/status/external-solr` -- Solr core admin ping
- `/status/external-fedora` -- rubydora connectivity ping

drh-stanford:
This PR fixes #48. It doesn't do *exactly* the same thing as the is_it_working implementation but it's close. Notably, the Fedora check now forces a GET from the Fedora server rather than using a cached value (that was internal to Rubydora gem).

NOTE that I wasn't able to test the `/status/external-solr` endpoint fully due to firewall restrictions. We'll need to deploy to stage to test that piece.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -20,7 +20,7 @@ OkComputer::Registry.register 'version', VersionCheck.new
 ##
 # EXTERNAL Services
 
-OkComputer::Registry.register 'external-solr', OkComputer::SolrCheck.new(Dor::SearchService.solr.uri.to_s)
+OkComputer::Registry.register 'external-solr', OkComputer::SolrCheck.new(Dor::SearchService.solr.uri.to_s.gsub(/\/$/, ''))
 
 # Simple check to ping the Fedora server by asking it to describe the repository
 class FedoraCheck < OkComputer::Check
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,0 +1,37 @@
+require 'okcomputer'
+
+OkComputer.mount_at = 'status' # use /status or /status/all or /status/<name-of-check>
+OkComputer.check_in_parallel = true
+OkComputer::Registry.deregister 'database' # we don't have a database
+
+##
+# REQUIRED checks
+
+# Simple echo of the VERSION file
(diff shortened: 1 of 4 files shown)


config/initializers/okcomputer.rb
```diff
@@ -0,0 +1,37 @@
...
+class FedoraCheck < OkComputer::Check
+  def check
+    conn = ActiveFedora::Base.connection_for_pid(0)
+    profile = conn.repository_profile # use vs. `profile` to force a GET call to the server for `/fedora/describe`
+    mark_message "Connected to #{profile['repositoryName']} #{profile['repositoryVersion']} via Rubydora #{Rubydora::VERSION}"
+  rescue => e
+    mark_failure
+    mark_message "Failure: #{e.class.name}: #{e.message}"
+  end
+end
```
jmartin-sul: would be nice to get a bit of feedback form @ndushay on this since she's written a bunch of okcomputer checks at this point.  i couldn't remember an app off-hand that uses both okcomputer and a fedora connection, so i mostly wanted to check that we didn't have some other standard way of doing this already out in the wild.

drh-stanford: a bit of history: the Fedora check was a built-in `rubydora` check in `is_it_working-cbeer` developed circa 2012. The above `FedoraCheck` is a simplified version that forces a GET call to `/fedora/describe` on each check.

@eefahy confirmed that we are not doing app-level Fedora checks via Nagios right now due to lack of Nagios support for certificate-based authentication (she's looking into that further). So, as far as I know, the `rubydora` checks are the only monitoring of Fedora that we do. We'd like to move to a situation where Nagios is doing direct monitoring of Fedora.

drh-stanford:
I've deployed this branch to stage and it's working as follows:

screen shot 2016-11-04 at 11 04 38 am

ndushay:
sorry to be late to the party.  I was fussing with a fedora check in okcomputer more akin to the one @cbeer did for is_it_working, but I never got it to work.  So this seems like a good stopgap approach to me.

PR #66, opened 2016-11-07
Add dlss_cops to bring code up to date with best practices

drh-stanford:
Adds the `dlss_cops` gem and runs `rubocop --auto-correct`. Any rubocop complaints that we're auto-corrected were placed into `.rubocop_todo.yml`.


cbeer:
Are we still using `dlss_cops`? Has rubycop integration between IDEs and external definitions gotten any better?

drh-stanford:
Ok, `dlss_cops` was what we were using on purl-fetcher -- see https://github.com/sul-dlss/purl-fetcher/blob/master/.rubocop.yml#L3

But we can do this later.

PR #70, opened 2016-11-08
Redirect the application root to the new okcomputer endpoint

cbeer:


config/routes.rb
```diff
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
-  root to: redirect('/is_it_working')
+  root to: redirect('/status/all')
```
drh-stanford: I think this should be `/status` or more specifically `/status/default`. @eefahy and I have not been using the `all` endpoint for monintoring thus far.

cbeer: I figured `/status/all` will provide us with the best overview of what's going on (and would be the equivalent to the `/is_it_working` route).

drh-stanford:
ok 👍

PR #72, opened 2016-11-08
unpin active-fedora.  bundle update dor-services to current latest (5.14.2)

jmartin-sul:
implicitly updates active-fedora to 8.2.1

@drh-stanford: are you ok with relaxing the AF restriction?  it seems like the issues we ran into that we know about have been resolved, and it'd be nice to get this upgraded to the degree we can, and out in the wild.


cbeer:
👏  while there's a team available to work on it (and a comfy safety net with honeybadger), getting this update out of the way early would be nice.

PR #74, opened 2016-11-09
Raise production log level to :info

### LINKED ISSUE #6: should prod log_level be debug?
is that the standard?  seems like a reasonable setting to launch with (and even keep if its not too noisy).  but i was mildly surprised it wasn't :info in the production settings.

cbeer:
Fixes #6

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -35,7 +35,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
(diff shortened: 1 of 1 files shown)


drh-stanford:
I'm wondering whether we should also disable the rack-timeout logging too.

```ruby
Rack::Timeout::Logger.disable
```

their logs look like so and it's roughly 1/3 of the information in the logs:

```
D, [2016-11-09T09:53:35.097445 #27336] DEBUG -- : source=rack-timeout id=31d6377ebe25b57d09a0533bb701cb7a timeout=60000ms service=4ms state=active
I, [2016-11-09T09:53:35.100746 #27336]  INFO -- : source=rack-timeout id=31d6377ebe25b57d09a0533bb701cb7a timeout=60000ms service=7ms state=completed
```

drh-stanford:
Our lock down of master seems to want the PR branch totally in sync with master. Github has an 'Update branch' button which I tried. It added a commit (6cce940) to this PR.

Issue #113, opened 2016-11-18
deploy on dor-indexing-app-dev


drh-stanford:
See SAR-3312

drh-stanford:
See SAR-3314

drh-stanford:
argo dev is getting the following error message when trying to connect to reindex on dor-indexing-app dev

```
[---] [2016-11-21 20:05:04.248] failed to reindex druid:zn732sn3193: SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed
```

cbeer:
Valid certificates are now in place.

PR #116, opened 2016-11-18
add dev deployment for Cap

drh-stanford:
This PR is connected to #113. It adds the `dev` environment to capistrano and does some cleanup too.


config/deploy/legacy.rb
```diff
@@ -1,6 +1,6 @@
 server "dor-indexing-app-legacy-prod.stanford.edu", user: fetch(:user), roles: %w{web app}
```
cbeer: This file can go away entirely.

PR #272, opened 2019-04-09
Avoid deprecated Dor::IndexingService.generate_index_logger

jcoyne:


ndushay:
coveralls ... why???

jcoyne:
@ndushay Coveralls went down a bit because I added two lines that are not covered.

PR #273, opened 2019-04-09
Move reindex_pid here from dor-services

jcoyne:
Ref https://github.com/sul-dlss/dor-services/issues/529


jcoyne:
@ndushay yes, it's already been deprecated in dor-services.

PR #274, opened 2019-04-09
Upgrade to rails 5.2

jcoyne:


config/environments/production.rb
```diff
@@ -25,17 +29,12 @@
...
-  # config.action_cable.url = 'wss://example.com/cable'
-  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
-
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :info
+  config.log_level = :debug
```
justinlittman: Debug?

jcoyne: This is the rails default and is intentional.

PR #291, opened 2019-07-02
update to latest dor-services

peetucket:


peetucket:
I'm not sure why it is claiming coverage decreased ... it is currently at 96% on master

jmartin-sul:
yeah, that's weird, but i agree it seems erroneous.  and i've def seen similarly erroneous coverage reports in the past, so whatever, no big deal.

jmartin-sul:
also the build report conflicts with the coveralls comment, so all the more 🤷‍♂

PR #309, opened 2019-09-24
fix how we pass optional params to rsolr

### LINKED ISSUE #297: When workflow changes occur, indexing of objects seems to lag
During accessioning, as robots operate on an object, the workflow steps are changing, potentially quickly, but the index of the object into solr seems to log a certain moment.  The impact is that user's don't think anything has happened yet, even though it has.  It would be good to:

- [x] Confirm what the actual typical lag in indexing is
- [x] Look at the indexing queuing infrastructure to see if there are any other changes that could be made

peetucket:
fixes #297 

The optional param of "commitWithin" was never making it to solr.

According to https://github.com/rsolr/rsolr/blob/master/README.rdoc under "Updating Solr" indicates we need to a pass a the params like this:

`solr.add documents, :add_attributes => {:commitWithin => 10}`

and not like we were doing like this:

`solr.add documents, {:commitWithin => 10}`

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -22,7 +22,7 @@ Metrics/AbcSize:
 # Configuration parameters: CountComments, ExcludedMethods.
 # ExcludedMethods: refine
 Metrics/BlockLength:
-  Max: 55
+  Max: 59
 
 # Offense count: 1
 # Configuration parameters: CountComments, ExcludedMethods.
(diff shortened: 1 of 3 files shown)


peetucket:
> @peetucket It would be ideal if you could add a test that would fail w/o the patch you added. If that's onerous, I'd understand not doing so.

I agree this would be nice.  Hopefully won't be hard with the mocking already setup.

peetucket:
added some tests to confirm new behavior

peetucket:
will fix rubocop in a bit!

peetucket:
build fixed

PR #310, opened 2019-09-24
Update rubocop dependency and configs

mjgiarlo:
Includes:
* Remove exclusions from RuboCop configuration
* Run rubocop auto-correct
* Use `rubocop-performance` and `rubocop-rails` gems and latest rubocop per recent practice
* Update rubocop and update rubocop config to better isolate exclusions


mjgiarlo:
Will work on fixing the build.

PR #320, opened 2019-11-25
Upgrade to dor-services-8

jcoyne:


jcoyne:
@andrewjbtw  is testing this on stage right now.

config/initializers/config.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Config.setup do |config|
+  config.const_name = 'Settings'
+  config.use_env = true
+  config.env_prefix = 'SETTINGS'
+  config.env_separator = '__'
+  # config.env_converter = :downcase
```
justinlittman: Cruft?

andrewjbtw:
For testing on stage, I've:

- registered
- accessioned
- versioned
- run some bulk actions 
- run reports
- clicked on and updated datastreams 
- clicked on and viewed workflows

I didn't see any issues with any of these checks.

config/initializers/config.rb
jcoyne: Not really.  This is something we usually do, but we have to convert all the Settings variables to lowercase first. Not something I want to add to this pr.

PR #327, opened 2020-01-07
Move indexing code into the app (from dor-services)

jcoyne:
This makes it possible to remove the indexing code from dor-services and see the one service responsible for writing to the index


ndushay:
that's a precipitous drop in coverage - are some spec files missing?

jcoyne:
@ndushay Fixed coverage drop.

PR #346, opened 2020-02-06
Upgrade to dor-services-9

jcoyne:


Gemfile
```diff
@@ -1,3 +1,4 @@
+
```
mjgiarlo: :fire: :keyboard: :fire:

app/indexers/workflows_indexer.rb
```diff
@@ -27,6 +27,6 @@ def workflows
   # TODO: remove Dor::Workflow::Document
   # @return [Workflow::Response::Workflows]
   def all_workflows
-    @all_workflows ||= Dor::Config.workflow.client.workflow_routes.all_workflows pid: resource.pid
```
mjgiarlo: :wave: `Dor::Config` :wave:

mjgiarlo:
@jcoyne do you think we should test this in stage before monday morning dep updates?

PR #349, opened 2020-02-15
Add a docker image

jcoyne:
As we don't want argo writing directly to solr, this will be required for testing argo


Gemfile
```diff
@@ -36,6 +36,7 @@ group :development, :test do
 end
 
 group :development do
+  gem 'listen', '~> 3.0.5'
```
jmartin-sul: what's the purpose of adding this gem?

jcoyne: It is required by our existing configuration: https://github.com/sul-dlss/dor_indexing_app/blob/b12e7a5ce02d9750ef27abb609bf2d66ece89e33/config/environments/development.rb#L40  I had the gem locally, so I didn't know it was a problem until building in a clean (docker) environement.

jmartin-sul: ah, got it, figured something like code reloading.  thanks for the explanation!

PR #358, opened 2020-03-03
Index administrative tags via DOR API

mjgiarlo:
Unblocks sul-dlss/dor-services-app#709
Part of sul-dlss/dor-services-app#679

As part of this work, stop indexing tags via the identity metadata datastream.

## Why was this change made?

To make sure tags added to objects via API are indexed.

No.

Yes, and it was tested successfully in stage and now there's an integration test: https://github.com/sul-dlss/infrastructure-integration-test/pull/34

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -36,6 +36,7 @@ Metrics/ParameterLists:
 # Configuration parameters: Max.
 RSpec/ExampleLength:
   Exclude:
+    - 'spec/indexers/administrative_tag_indexer_spec.rb'
     - 'spec/indexers/composite_indexer_spec.rb'
     - 'spec/indexers/describable_indexer_spec.rb'
     - 'spec/indexers/identifiable_indexer_spec.rb'
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem 'rack-timeout', '~> 0.5.1'
 gem 'bootsnap', '>= 1.1.0', require: false
 
 gem 'dor-services', '~> 9.0', '>= 9.2.1' # must be 9.2.1 to get fix from https://github.com/sul-dlss/dor-services/pull/695
-gem 'dor-services-client', '~> 4.11'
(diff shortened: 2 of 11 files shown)


jcoyne:
Currently this work is performed by `IdentityMetadata#to_solr`, via DataIndexer.  Do we need to turn that code off?

mjgiarlo:
@jcoyne :speech_balloon: 
> Currently this work is performed by `IdentityMetadata#to_solr`, via DataIndexer. Do we need to turn that code off?

This PR now does that, thanks to the work you've done. Thanks!

jcoyne:
So this is currently blocked by https://github.com/sul-dlss/dor-services-client/pull/154 and a release of DSC, right?

mjgiarlo:
@jcoyne :speech_balloon: 
> So this is currently blocked by [sul-dlss/dor-services-client#154](https://github.com/sul-dlss/dor-services-client/pull/154) and a release of DSC, right?

Yes, this is blocked by the DSC work being merged and cut in a release. I am testing all the components now to make sure they work as expected, and then I will unhold the DSC PR.

Dockerfile
```diff
@@ -5,6 +5,7 @@ FROM ruby:2.7.0-alpine
 
 RUN apk add --update --no-cache  \
   build-base \
+  git \
```
jcoyne: is this necessary?

mjgiarlo: @jcoyne It was when I was pointing DIA at a GitHub branch. I figured I'd leave it in to make it easier to do this in the future.

jcoyne: But this makes the images substantially bigger, which means pulling this image for testing takes longer.

mjgiarlo: @jcoyne looks like it adds ~16MB to a package that is ~155MB. :man_shrugging: I can remove it.

jcoyne: 10% reduction!

mjgiarlo: Removed

PR #360, opened 2020-03-05
Add an indexer for each datastream

jcoyne:
## Why was this change made?

This gives us a place to move code to as we move the indexing logic out of the datastream models to prepare to migrate away from Fedora 3

n/a


justinlittman:
This is much clearer than indexing magic.

PR #371, opened 2020-03-13
[HOLD] Pass the cocina instance to each indexer

jcoyne:
*Note* This requires that every object has a sourceId (or dor-services-app can't create a Cocina model).  Currently not everything has one.  All data needs to be remediated before this can go forward.

## Why was this change made?
This makes it possible to migrate each indexer to use the cocina model

n/a

no

### DIFF
--- a/app/services/indexer.rb
+++ b/app/services/indexer.rb
@@ -94,7 +94,7 @@ class Indexer
     Dor::Set => SET_INDEXER
   }.freeze
 
-  def self.for(obj)
-    INDEXERS.fetch(obj.class).new(resource: obj)
+  def self.for(fedora:, cocina:)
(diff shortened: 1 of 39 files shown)


app/services/indexer.rb
```diff
@@ -94,7 +94,7 @@ class Indexer
     Dor::Set => SET_INDEXER
   }.freeze
 
-  def self.for(obj)
-    INDEXERS.fetch(obj.class).new(resource: obj)
+  def self.for(fedora:, cocina:)
```
justinlittman: Call `resource` instead of `fedora` for consistency?

jcoyne:
Superseded by #442

PR #372, opened 2020-03-13
Revert changes to ETD handling

jcoyne:
## Why was this change made?

When we changed etds to use the ETD_INDEXER, we lost workflows and datastream displays.
This reverts those commits so we treat ETDs as Dor::Item and removes the unused ETD_INDEXER since it was never used.

n/a

no.


justinlittman:
So there are no ETD unique properties that need to be indexed?

jcoyne:
@justinlittman I don't believe so.

peetucket:
Wasn't there some issue about the associated APO not being indexed for ETDs that this change was originally attended to address?

jcoyne:
@ndushay the shas that were reverted are in the commit messages.

peetucket:
looks fine

PR #373, opened 2020-03-16
Update dependencies

sul-devops-team:


spec/spec_helper.rb
```diff
@@ -90,7 +90,7 @@
   # Print the 10 slowest examples and example groups at the
   # end of the spec run, to help surface which specs are running
   # particularly slow.
-  config.profile_examples = 10
+  # config.profile_examples = 10
```
aaron-collier: Is this change intentional and/or necessary?

ndushay: it's not necessary, it is intentional.  IMO, there is no need to profile stuff unless the tests run slow, and it clutters up the test results in my terminal.  Especially when fixing bugs.

PR #374, opened 2020-03-18
Convert settings values to lowercase

mjgiarlo:
## Why was this change made?

This provides consistency with other projects and uses the config gem recommended style. I added this while working on https://github.com/sul-dlss/dor_indexing_app/pull/358.

Accompanying shared_configs PRs will be created momentarily.

Only in dockerland (dev/test), not in deployed environments.

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -26,9 +26,9 @@ class DorController < ApplicationController
   # Returns a Logger instance for recording info about indexing attempts
   # @param [String] entry_id an extra identifier for the log entry.
   def generate_index_logger(entry_id)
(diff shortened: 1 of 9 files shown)


jcoyne:
Do you want to do this where we have an upper case & lower case version?  Then we wouldn't need to coordinate this deploy with the code deploy.  Note, this is not a demand for change, just wondering about strategy.

mjgiarlo:
@jcoyne :speech_balloon: 
> Do you want to do this where we have an upper case & lower case version? Then we wouldn't need to coordinate this deploy with the code deploy. Note, this is not a demand for change, just wondering about strategy.

The three shared_configs PRs are now up, FWIW.

PR #400, opened 2020-06-02
Write and validate OpenAPI specification

mjgiarlo:
## Why was this change made?

Consistency with other projects, and because DIA's API is super small, so why not?

Unit tests, and tested in the staging environment. (All integration tests passed.)

README


mjgiarlo:
@ndushay Not going to deploy this to prod... but, as FR, do you otherwise have concerns or questions about this before it is merged?

PR #409, opened 2020-06-05
Remove ds_specs_ssim

jcoyne:
## Why was this change made?
This is no longer necessary since https://github.com/sul-dlss/argo/pull/2103 and it was causing significant slowness of the indexing process, especially when on external datastreams. This example takes over 20s:
```ruby
druid = 'druid:fw763ds3827'
resource = Dor.find(druid)
resource.datastreams['workflows'].datastream_spec_string
```

Test suite

n/a


ndushay:
merging, but feeling like I'm flying blind, a bit.  I assume theses references to `ds_specs` are nothing to do with this.

https://github.com/search?q=org%3Asul-dlss+ds_specs&type=Code

PR #411, opened 2020-06-05
Only index the tags we use

jcoyne:
## Why was this change made?
This keeps the index size under control and helps us better understand what data we depend on.

Test suite

n/a


jcoyne:
@justinlittman because of analysis of Argo.

PR #414, opened 2020-06-09
Extract fields to a lookup table

jcoyne:
## Why was this change made?

Fixes too many parameters passed to a method

Test suite

n/a


jcoyne:
@justinlittman I rather disagree. This clearly groups together the various fields that the class can set, rather than having them sprinkled throughout the logic.  I think it's much clearer where the methods are invoked why the parameters are being passed.

PR #418, opened 2020-06-26
Stop indexing ds_spec_ssim

jcoyne:
## Why was this change made?
This is no longer used as of https://github.com/sul-dlss/argo/pull/2103 and it was causing significant slowness of the indexing process, especially when on external datastreams. This example takes over 20s:
```ruby
druid = 'druid:fw763ds3827'
resource = Dor.find(druid)
resource.datastreams['workflows'].datastream_spec_string
```

test suite

n/a


jcoyne:
Duplicate of #409

PR #442, opened 2020-10-08
Optionally support cocina lookup

jcoyne:
## Why was this change made?

This will allow us to use cocina model if it exists.  Using dry-monads to show that we have to check for its value until we have consitent conformant data

Tested on stage.


mjgiarlo (approved):
LGTM and I understand this has been getting tested on stage

PR #462, opened 2021-01-13
Add debug logging for development

jcoyne:
And avoid custom loggers.

## Why was this change made?

This is helpful for diagnosing where a slowdown is happening.
There should be a single centralized log so that requests can easily be traced. See https://12factor.net/logs

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -5,7 +5,7 @@ class DorController < ApplicationController
   include Dry::Monads[:result]
 
   def reindex
-    @solr_doc = reindex_pid params[:pid], logger: generate_index_logger(request.uuid), add_attributes: { commitWithin: params.fetch(:commitWithin, 1000).to_i }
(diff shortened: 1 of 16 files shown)


app/controllers/dor_controller.rb
```diff
@@ -52,11 +40,13 @@ def reindex_pid(pid, logger:, add_attributes:)
         Failure(:conversion_error)
       end
     end.format('%n realtime %rs total CPU %ts').gsub(/[()]/, '')
-
+    logger.info "document found, now generating document solr"
```
justinlittman, line 43: Here and L48: Should these be `Rails.logger`?

justinlittman (approved):
Approve pending consideration of comment.

Do you intend to merge into master or just deploy the branch?

app/controllers/dor_controller.rb
jcoyne: the `logger` method is already on all controllers.

mjgiarlo (approved):

PR #483, opened 2021-02-27
Convert content metadata indexer to completely decouple from Fedora.

jcoyne:
Stop indexing resource counts and resource_types_ssim

## Why was this change made?
The individual count fields are not used, but the resouce_types_ssim is


cbeer:
Besides being unused, is there a reason to get rid of this field? It's been really useful in the past to track down fixture objects.

jcoyne:
@cbeer it's just to simplify before decoupling from fedora 3. Can you explain more about how you would use this to track down a fixture object?

mjgiarlo (approved):

mjgiarlo (approved):

PR #488, opened 2021-03-02
Index the descMetadata using cocina

jcoyne:
## Why was this change made?

Decouple from Fedora 3

### DIFF
--- a/app/indexers/descriptive_metadata_datastream_indexer.rb
+++ b/app/indexers/descriptive_metadata_datastream_indexer.rb
@@ -1,14 +1,18 @@
 # frozen_string_literal: true
 
 class DescriptiveMetadataDatastreamIndexer
-  attr_reader :resource
+  attr_reader :cocina
 
   def initialize(resource:, cocina:)
-    @resource = resource
+    @cocina = cocina
   end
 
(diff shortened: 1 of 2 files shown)


app/indexers/descriptive_metadata_datastream_indexer.rb
```diff
@@ -1,14 +1,18 @@
 # frozen_string_literal: true
 
 class DescriptiveMetadataDatastreamIndexer
-  attr_reader :resource
+  attr_reader :cocina
 
   def initialize(resource:, cocina:)
-    @resource = resource
+    @cocina = cocina.value!
```
justinlittman, line 7: Why is a Success being passed instead of cocina?

jcoyne: Success is a wrapper around the cocina, because not all objects have a cocina.  However after I rebase, this will be irrelevant.

justinlittman (approved):

app/indexers/descriptive_metadata_datastream_indexer.rb
justinlittman: Irrelevant seems less confusing to me.

PR #504, opened 2021-03-03
Handle the case when an object without identification sub-schema

jcoyne:
## Why was this change made?
A null pointer error was raised when testing in QA.

### DIFF
--- a/spec/indexers/identifiable_indexer_spec.rb
+++ b/spec/indexers/identifiable_indexer_spec.rb
@@ -75,8 +75,6 @@ RSpec.describe IdentifiableIndexer do
   describe '#to_solr' do
     let(:doc) { indexer.to_solr }
     let(:mock_rel_druid) { 'druid:qf999gg9999' }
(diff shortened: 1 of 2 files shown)


spec/indexers/identifiable_indexer_spec.rb
```diff
@@ -148,5 +146,32 @@
         expect(doc).to match a_hash_including('metadata_source_ssi' => 'DOR')
       end
     end
+
+    context 'with identification sub-schema' do
```
mjgiarlo, line 150: Should this be `with no identification sub-schema`?

mjgiarlo (approved):
LGTM modulo the one context label tweak.

spec/indexers/identifiable_indexer_spec.rb
jcoyne: Yes, thank you.

mjgiarlo (approved):

PR #532, opened 2021-03-12
Update to cocina-models-0.55

jcoyne:


Gemfile
```diff
@@ -17,6 +17,7 @@ gem 'rack-timeout', '~> 0.5.1'
 
 # Reduces boot times through caching; required in config/boot.rb
 gem 'bootsnap', '>= 1.1.0', require: false
+gem 'cocina-models', github: 'sul-dlss/cocina-models', branch: 'main' # '~> 0.54.0'
```
justinlittman, line 20: cruft?

justinlittman (changes requested):

Gemfile
jcoyne: This was before 0.55 was released. I'm going to update now.

justinlittman (approved):

PR #546, opened 2021-03-24
embargo_metadata_indexer uses cocina data instead of active-fedora

### LINKED ISSUE #508: Decouple EmbargoMetadataDatastreamIndexer from ActiveFedora
We must resolve https://github.com/sul-dlss/dor-services-app/issues/2414 first

ndushay:
~~possibly the cause of a flurry of alerts on -qa~~

This has been tested thoroughly by some non-default infrastructure-integration-tests for embargo (for ETD, hydrus, and h2).   Once this is merged, look for a followup PR in infrastructure-integration-tests.

## Why was this change made?

Fixes #508

more decoupling from Fedora3 - don't use it to index embargo information

unit tests and a new integration test (see sul-dlss/infrastructure-integration-test/pull/217) to test it in situ.

### DIFF
--- a/app/indexers/embargo_metadata_indexer.rb
+++ b/app/indexers/embargo_metadata_indexer.rb
@@ -12,9 +12,9 @@ class EmbargoMetadataIndexer
   def to_solr
     {}.tap do |solr_doc|
       embargo_release_date = embargo_release_date(cocina)
-      if future?(embargo_release_date)
+      if embargo_release_date.present?
         solr_doc['embargo_status_ssim'] = ['embargoed']
-        solr_doc['embargo_release_dtsim'] = [embargo_release_date.utc.strftime('%FT%TZ')]
+        solr_doc['embargo_release_dtsim'] = [embargo_release_date.utc.iso8601]
       end
     end
   end
@@ -23,11 +23,5 @@ class EmbargoMetadataIndexer
 
   def embargo_release_date(cocina)
     cocina.access.embargo.releaseDate if cocina.access.embargo&.releaseDate.present?
-  end
-
-  def future?(embargo_release_date)
-    return false unless embargo_release_date
-
-    embargo_release_date > DateTime.now
   end
 end
--- a/app/indexers/embargo_metadata_indexer.rb
+++ b/app/indexers/embargo_metadata_indexer.rb
@@ -1,27 +1,33 @@
 # frozen_string_literal: true
 
 class EmbargoMetadataIndexer
-  attr_reader :resource
+  attr_reader :cocina
 
-  def initialize(resource:, **)
-    @resource = resource
+  def initialize(cocina:, **)
+    @cocina = cocina
   end
 
   # These fields are used by the EmbargoReleaseService in dor-services-app
   # @return [Hash] the partial solr document for embargoMetadata
   def to_solr
-    {
-      'embargo_status_ssim' => embargo_status
-    }.tap do |solr_doc|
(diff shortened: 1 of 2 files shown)


app/indexers/embargo_metadata_indexer.rb
```diff
@@ -1,27 +1,33 @@
...
 
-  # rubocop:disable Lint/UselessAccessModifier
   private
 
-  # rubocop:enable Lint/UselessAccessModifier
+  def embargo_release_date(cocina)
+    cocina.access.embargo.releaseDate if cocina.access.embargo&.releaseDate.present?
+  end
+
+  def future?(embargo_release_date)
```
jcoyne, line 28: I think this check could potentially be harmful, because dor-services-app relies on this index so it can find any items that have a past embargo and then lift them.https://github.com/sul-dlss/dor-services-app/blob/main/app/services/embargo_release_service.rb#L8

jcoyne (changes requested):

app/indexers/embargo_metadata_indexer.rb
ndushay: ok - so your recommendation is to put them in the index whenever they are present?

jcoyne: Yes.

jcoyne: And there will be a dependency on this change: https://github.com/sul-dlss/dor-services-app/pull/2589

ndushay: ok.  the discussion ticket is probably why I wasn't indexing the ones in the past in any case.

ndushay: the change you asked for here ... means I will index any embargo data found in cocina, for future or past embargos.   So sul-dlss/dor-services-app#2589 is irrelevant to this PR.

jcoyne: The change you have here says, "any embargo that is past has been released" and that's not a true statement.  An embargo gets released when it gets released, and then (after sul-dlss/dor-services-app#2589), it will no longer show up in the record.

ndushay: fixed - all embargo dates encountered are now treated equally.  We have a zero tolerance policy against discrimination.

jcoyne (approved):

app/indexers/embargo_metadata_indexer.rb
```diff
@@ -1,27 +1,27 @@
...
   def to_solr
-    {
-      'embargo_status_ssim' => embargo_status
-    }.tap do |solr_doc|
-      solr_doc['embargo_release_dtsim'] = Array(release_date.first.utc.strftime('%FT%TZ')) if release_date.first.present?
+    {}.tap do |solr_doc|
+      embargo_release_date = embargo_release_date(cocina)
+      if embargo_release_date.present?
+        solr_doc['embargo_status_ssim'] = ['embargoed']
+        solr_doc['embargo_release_dtsim'] = [embargo_release_date.utc.strftime('%FT%TZ')]
```
jcoyne, line 17: 💨 I think you can do `embargo_release_date.utc.iso8601` here and it is equivalent and maybe easier to understand. 🤷

ndushay: changed - thanks.  I thought I ran into trouble with the milliseconds included in iso8601 but apparently I misremembered.

aaron-collier (approved):

PR #552, opened 2021-03-25
Support the 'Author' role when indexing authors

jcoyne:
## Why was this change made?

And allow names with basic values.

Fixes the integration test failure for hydrus

Integration test.


ndushay (approved):
I approve, but I fixed the rubocop errors and sort of paired, so maybe I don't count.

jmartin-sul (approved):

PR #573, opened 2021-03-31
Index the model for items that fail validation

jcoyne:
## Why was this change made?

Lots of items are failing on qa/stage with:
```
Solr document (id: druid:nb022qg2431) is missing required has_model_ssim field.
```


mjgiarlo:
@jcoyne should we test this in a deployed env or meh?

jcoyne:
@mjgiarlo it is currently running in qa and fixed all the problems we saw.

Issue #575, opened 2021-04-01
fix originInfo_date_created_tesim indexing 


ndushay:
### What Needs to be Done

In `DescriptiveMetadataIndexer`, `originInfo_date_created_tesim` is not being populated correctly from cocina.

Finding the correct creation date in Cocina is at least partially expressed in PR #553, and will be more completely described by @arcadiafalcone either on ticket #571 (and hopefully, maybe here as well?)
- Ensure we account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g.https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/origin_info_date_spec.rb. That is, if the code only expects "date.value" it won't find any values inside a structuredValue, a parallelValue (i.e. multi-lingual), and all that sort of stuff.

### User Impact

- we stopped displaying the place of publication in Argo search result titles. Same for record views. And probably other places where we present citation info.
- there is a failing integration test due to this (the one where we deposit an item via Hydrus)

ndushay:
@arcadiafalcone this is partially implemented in PR #575 ... to the best of my ability, anyway.

ndushay:
This is now done.  Yay!

PR #579, opened 2021-04-01
combine describable_indexer into descriptive_metadata_indexer

ndushay:
## Why was this change made?

There was no discernable reason for there to be two separate classes for the same thing, so I moved the contents of "DescribableIndexer" into the existing DescriptiveMetadataIndexer

This will also help with issue #575 and some other tickets.

All the tests previously written still pass (and some were moved to use rspec naming conventions)


app/indexers/descriptive_metadata_indexer.rb
```diff
@@ -1,46 +1,195 @@
...
+# rubocop:disable Metrics/ClassLength
 class DescriptiveMetadataIndexer
   attr_reader :cocina
 
   def initialize(cocina:, **)
     @cocina = cocina
   end
 
-  # @return [Hash] the partial solr document for descMetadata
-  # TODO: Naomi will be writing issues that get correct mapping from Arcadia, that
```
ndushay, line 11: These issues have been written.

mjgiarlo (approved):

PR #580, opened 2021-04-02
improve originInfo_date_created_tesim indexing

ndushay:
## Why was this change made?

Fixes sul-dlss/infrastructure-integration-test/issues/219  - broken hydrus integration test 

Also a partial fix for #575 - my best guess of the algorithm Arcadia will provide.

test suite of this app and then running integration tests.

### DIFF
--- a/spec/services/indexer_spec.rb
+++ b/spec/services/indexer_spec.rb
@@ -124,7 +124,8 @@ RSpec.describe Indexer do
                       'status' => 'primary',
                       'encoding' => {
                         'code' => 'w3cdtf'
-                      }
+                      },
+                      'type' => 'creation'
                     }
                   ]
                 },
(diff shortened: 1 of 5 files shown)


spec/services/indexer_spec.rb
```diff
@@ -162,7 +163,7 @@
         it 'has required fields' do
           expect(solr_doc).to include('milestones_ssim', 'released_to_ssim', 'wf_ssim', 'tag_ssim')
 
-          expect(solr_doc['originInfo_date_created_tesim']).to eq ['2021-01-01']
```
jcoyne, line 165: I'm confused about why this expectation is changing? Doesn't the integration test expect YYYY-MM-DD?

ndushay: thanks for spotting this -- the publication date is a year (it's a facet) while this date is ... a date.  :-)   Fixed.

jcoyne (approved):

PR #582, opened 2021-04-02
Add spec for origin_info_date_created_tesim

### LINKED ISSUE #575: fix originInfo_date_created_tesim indexing

arcadiafalcone:
## Why was this change made?
Part of #575

CircleCI

n/a


ndushay (changes requested):
@arcadiafalcone I think this is more a "create_date" spec, no?  or even origin_info_date_created  spec?   (🤮 ) Maybe change the file name?

ndushay (approved):

PR #586, opened 2021-04-05
descriptive_meatdata_indexer_spec: switch to ruby hashes

ndushay:
## Why was this change made?

For the same reasons we prefer ruby hashes in the dor-services-app for mapping specs -- the syntax highlighting in our IDEs make it a lot easier to spot errors.

unit tests


mjgiarlo (approved):

jcoyne:
🥩 data!

PR #589, opened 2021-04-06
Fix genre indexing.

### LINKED ISSUE #566: fix genre indexing
### Background

This ticket is to make the `sw_genre_ssim` field contain values as much like the ones we were getting from the stanford-mods method as is reasonable.

Per PR #517, `sw_genre_ssim` Solr field was formerly populated by calling `sw_genre` in stanford-mods (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/searchworks.rb#L305).  

"term_value" is from the "om" gem approach to metadata mapping from a bazillion years ago with Hydra (yes, it was Hydra then).  It is defined in the mods gem:  https://github.com/sul-dlss/mods/blob/master/lib/mods/nom_terminology.rb#L72.  It then seems to normalize what it finds with particular capitalizations.

I'm not sure how well the genre terms from stanford-mods correspond to the terms we use in Cocina -- You can look at what is in the SearchWorks genre facet https://searchworks.stanford.edu/?f%5Baccess_facet%5D%5B%5D=At+the+Library (that should cover most all of them).  You can also look at the mappings from MODS to Cocina here to get a sense of what wrinkles might be present in the Cocina genre data:   https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/genre_spec.rb.  

    - ask her how closely the Cocina genre information maps to what was in MODS; ask her and Andrew if Cocina data is more or less ok as is;  ask them what capitalization is expected, if it's a controlled vocabulary, etc. 
- ensure you account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g. https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/genre_spec.rb.  

### User Impact
We are likely giving Argo users a mistaken impression of how their genre data will appear in SearchWorks.

justinlittman:
closes #566

## Why was this change made?
We didn't take into account all of the cocina descriptive intricacies.

Unit

NA

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -80,7 +80,7 @@ class DescriptiveMetadataIndexer
   end
 
   def genres
-    @genres ||= forms.select { |form| form.type == 'genre' }.map(&:value)
+    @genres ||= forms.flat_map { |form| form.parallelValue || form }.select { |form| form.type == 'genre' }.map(&:value)
   end
 
   FORMAT = {
--- a/spec/indexers/descriptive_metadata/genre_spec.rb
+++ b/spec/indexers/descriptive_metadata/genre_spec.rb
@@ -90,7 +90,7 @@ RSpec.describe DescriptiveMetadataIndexer do
         }
       end
 
-      xit 'uses genre value' do
+      it 'uses genre value' do
         expect(doc).to include('sw_genre_ssim' => ['photographs'])
       end
     end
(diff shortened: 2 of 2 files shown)


justinlittman:
@arcadiafalcone I made some changes in the specs to reflect the existing code. Can you review?

jcoyne (approved):

arcadiafalcone (approved):

jcoyne:
@arcadiafalcone @justinlittman sorry, I missed the message to Arcadia to review.

PR #593, opened 2021-04-08
fix title indexing

### LINKED ISSUE #563: fix title indexing (spec'ed but not done)
### Background

### What Needs to be Done

This ticket is to make the `sw_display_title_tesim` field contain values as much like the ones we were getting from the stanford-mods method as is reasonable.

Per PR #517, `sw_display_title_tesim` Solr field was formerly populated by calling `sw_title_display` in stanford-mods (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/searchworks.rb#L193).  Tracing through the logic, you see something like this:

`sw_title_display` gets a value from `sw_full_title` and then removes specific trailing punctuation
`sw_full_title` gets a value from non-sorting title chars, then title, then " : " then subTitle and then does stuff with partName and partNumber if present ... and then does some stuff with trailing punctuation.

    - ask her what to do if there are mulitple title values in Cocina
- Write tests that exercise different use cases
- ensure you account for the possibility of parallel data (e.g. multilingual) and structured data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g. https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/title_info_spec.rb.    That is, if the code only expects "title.value" it won't finded any structuredValues or parallelValues, and all that good stuff.

### User Impact
We are likely not populating the title field for some use cases (e.g. when the data is multi-lingual, or when it is structured) and we are likely giving Argo users a mistaken impression of how their data will appear in SearchWorks.

ndushay:
## Why was this change made?

Fixes #563

Fixes indexing for `sw_display_title_tesim` Solr field per Arcadia's specs

unit tests; additionally ran some integration tests after deploying to qa

### DIFF
--- a/app/services/title_builder.rb
+++ b/app/services/title_builder.rb
@@ -4,14 +4,15 @@ class TitleBuilder
   # @param [Array<Cocina::Models::Title>] titles
   # @returns [String] the title value for Solr
   def self.build(titles)
-    title = titles.first
+    title = primary_title(titles) || titles.first
     if title.value
-      title.value
+      my_title = title.value
     elsif title.structuredValue
-      build_structured(title.structuredValue)
+      my_title = build_structured(title.structuredValue)
     elsif title.parallelValue
-      build(title.parallelValue)
+      my_title = build(title.parallelValue)
     end
+    remove_trailing_punctuation(my_title.strip).strip if my_title.present?
   end
 
   def self.build_structured(titles)
@@ -25,6 +26,7 @@ class TitleBuilder
     last = title_parts.pop
     title_parts.join + last.first # Drops the join token from the last element
   end
+  private_class_method :build_structured
 
   def self.join_token(type)
     case type
@@ -38,4 +40,27 @@ class TitleBuilder
       ' : '
     end
   end
+  private_class_method :join_token
+
+  def self.remove_trailing_punctuation(title)
+    title.gsub(/[\.,;:\/\\]$/, '')
+  end
+  private_class_method :remove_trailing_punctuation
+
+  # @param [Array<Cocina::Models::Title>] titles
+  # @reaturn [Cocina::Models::Title, nil] title that has status=primary
+  def self.primary_title(titles)
+    primary_title = titles.find do |title|
+      title.status == 'primary'
+    end
+    return primary_title if primary_title.present?
+
+    parallel_title_primary = titles.find do |title|
+      title.parallelValue&.find do |parallel_title|
+        parallel_title.status == 'primary'
+      end
+    end
+    parallel_title_primary
+  end
+  private_class_method :primary_title
 end
--- a/app/services/title_builder.rb
+++ b/app/services/title_builder.rb
@@ -4,24 +4,28 @@ class TitleBuilder
   # @param [Array<Cocina::Models::Title>] titles
   # @returns [String] the title value for Solr
   def self.build(titles)
-    title = primary_title(titles) || first_untyped_title(titles) || titles.first
-    if title.value
-      my_title = title.value
-    elsif title.structuredValue
-      my_title = title_from_structured_values(title.structuredValue, non_sorting_char_count(title))
-    elsif title.parallelValue
-      my_title = build(title.parallelValue)
-    end
-    remove_trailing_punctuation(my_title.strip) if my_title.present?
+    cocina_title = primary_title(titles) || first_untyped_title(titles) || titles.first
+    result = if cocina_title.value
+               cocina_title.value
+             elsif cocina_title.structuredValue
+               title_from_structured_values(cocina_title.structuredValue, non_sorting_char_count(cocina_title))
+             elsif cocina_title.parallelValue
+               build(cocina_title.parallelValue)
+             end
+    remove_trailing_punctuation(result.strip) if result.present?
   end
 
   # rubocop:disable Metrics/BlockLength
   # rubocop:disable Metrics/CyclomaticComplexity
   # rubocop:disable Metrics/MethodLength
   # rubocop:disable Metrics/PerceivedComplexity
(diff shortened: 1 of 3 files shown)


app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
   # @param [Array<Cocina::Models::Title>] titles
   # @returns [String] the title value for Solr
   def self.build(titles)
-    title = titles.first
+    title = primary_title(titles) || first_untyped_title(titles) || titles.first
     if title.value
```
jcoyne, line 8: Can you pull the variable assignment out of the conditional, e.g:
```ruby
my_title = if title.value
  title.value
elsif title.structuredValue
  title_from_structured_values(...)
else
...
end
```

app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
...
+  #   respect order of occurrence
+  #   separated from each other by comma space
+  # rubocop:disable Metrics/PerceivedComplexity
+  def self.title_from_structured_part(structured_values)
+    title_from_part = ''
+    structured_values.each do |structured_value|
+      case structured_value.type&.downcase
+      when 'part name', 'part number'
+        title_from_part = if title_from_part&.strip.present?
+                            "#{title_from_part.sub(/[ .,]*$/, '')}, #{structured_value.value&.strip}"
```
jcoyne, line 124: why is `value&.strip` used here? Are you expecting it to be nil?

app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
...
+  # combine part name and part number:
+  #   respect order of occurrence
+  #   separated from each other by comma space
+  # rubocop:disable Metrics/PerceivedComplexity
+  def self.title_from_structured_part(structured_values)
+    title_from_part = ''
+    structured_values.each do |structured_value|
+      case structured_value.type&.downcase
+      when 'part name', 'part number'
+        title_from_part = if title_from_part&.strip.present?
```
jcoyne, line 123: title_from_part can never be nil here, so why use the `&.`?

app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
...
+    end
+  end
+  private_class_method :primary_title
+
+  # @param [Array<Cocina::Models::Title>] titles
+  # @return [Cocina::Models::Title, nil] first title that has no type attribute
+  def self.first_untyped_title(titles)
+    titles.find do |title|
+      if title.parallelValue.present?
+        title.parallelValue&.find { |parallel_value| parallel_value.type.nil? }
```
jcoyne, line 98: you just checked that parallel value was present, so why use `parallelValue&.find`?

app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
...
+    structured_values.each do |structured_value|
+      # There can be a structuredValue inside a structuredValue.  For example,
+      #   a uniform title where both the name and the title have internal StructuredValue
+      return title_from_structured_values(structured_value.structuredValue, non_sorting_char_count) if structured_value.structuredValue
+
+      value = structured_value.value&.strip
+      next unless value
+
+      # additional types:  name, uniform ...
+      case structured_value.type&.downcase
```
jcoyne, line 34: the `type&.downcase` implies that this might be `nil`, but there is no when or else case for that value. Is that something we need to deal with?

app/services/title_builder.rb
```diff
@@ -4,38 +4,130 @@ class TitleBuilder
...
-      if title.structuredValue
-        [build_structured(title.structuredValue), '']
+  # rubocop:disable Metrics/BlockLength
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/PerceivedComplexity
+  def self.title_from_structured_values(structured_values, non_sorting_char_count)
+    structured_title = ''
+    title_from_part = ''
+    structured_values.each do |structured_value|
```
jcoyne, line 25: This appears to overwrite `structured_title` every time through this loop and then the method returns whatever one was set last?  Is that right?

ndushay: there is only one ultimate value for each structuredValue -- it is the amalgamation of all the diff pieces.

app/services/title_builder.rb
ndushay: I don't know;  Arcadia didn't have an example for it.  I asked her what other types will show up here and she is looking into it.  No reason to hold up this PR for those edge cases.

app/services/title_builder.rb
ndushay: good point; removing.

app/services/title_builder.rb
ndushay: good point; removing.

app/services/title_builder.rb
ndushay: just in case.  I've found that data is often dirty.

app/services/title_builder.rb
ndushay: good idea; done.

ndushay (commented):
thanks for the review;  i took nearly all your suggestions and the code is better for them.  Ready for re-reivew.

app/services/title_builder.rb
aaron-collier: this looks like it could result in a "title_part_value, " (with comma blank at the end) if the structured value is nil. Is that ok?

ndushay: any trailing punct is going to be removed with remove_trailing_punctuation as the last line of the public build method

aaron-collier (approved):

PR #594, opened 2021-04-08
title_builder: improve a yard comment

ndushay:
## Why was this change made?

to clarify a method

### DIFF
--- a/app/services/title_builder.rb
+++ b/app/services/title_builder.rb
@@ -2,7 +2,7 @@
 
 class TitleBuilder
   # @param [Array<Cocina::Models::Title>] titles
-  # @returns [String] the title value for Solr
+  # @return [String] the title value for Solr
   def self.build(titles)
     cocina_title = primary_title(titles) || first_untyped_title(titles) || titles.first
     result = if cocina_title.value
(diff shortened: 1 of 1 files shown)


aaron-collier (approved):

app/services/title_builder.rb
```diff
@@ -21,7 +21,8 @@ def self.build(titles)
   # rubocop:disable Metrics/PerceivedComplexity
   # @param [Array<Cocina::Models::StructuredValue>] structured_values - the individual pieces of a structuredValue to be combined
   # @param [Integer] the length of the non_sorting_characters
-  # @returns [String] the title value from combining the pieces of the structured_values according to type and order of occurrence
+  # @returns [String] the title value from combining the pieces of the structured_values according to type and order of occurrence,
```
mjgiarlo, line 24: It looks like 6 of the 41 uses of `@return(s)` in DIA use `@returns` instead of `@return` and I believe the latter is more correct per YARD documentation. Would you consider changing `@returns` to `@return` here? Alternatively, I'd be happy to put in a follow-up PR to replace/fix all of these for consistency and increased C O R R E C T N E S S. Up to you!

ndushay: i will fix.

mjgiarlo (approved):

PR #604, opened 2021-04-13
fix language indexing.

### LINKED ISSUE #565: fix language indexing
### Background

### What Needs to be Done

This ticket is to make the `sw_language_ssim` field contain values as much like the ones we were getting from the stanford-mods method as is reasonable.

Per PR #517, `sw_language_ssim` Solr field was formerly populated by calling `sw_language_facet` in stanford-mods (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/searchworks.rb#L24).  Tracing through the logic,  you see it looks for language codes (in our case, ask Arcadia which authority values for lang codes) and has a hash to look up the codes https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/searchworks_languages.rb.  We could include the gem for this hash lookup, or perhaps Arcadia would like us to use the actual authority text value (which is truthfully where that hash was populated from originally).  Note that it also uses text values as the Solr field allows for multiple values.

    - ask her what to do if there are mulitple authorities or codes and text values, multiple scripts, etc. in Cocina
- ensure you account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g. https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/language_spec.rb.    That is, if the code only expects "language.value" it won't find any values inside a script block, etc. and all that sort of stuff.

### User Impact
We are likely not populating the language Solr fields for some use cases (e.g. when the data has a script) and we are likely giving Argo users a mistaken impression of how their data will appear in SearchWorks.

### Completion

Part of completing this issue is addressing all of the `xit` specs in spec/indexers/descriptive_metadata/language_spec.rb

justinlittman:
closes #565

## Why was this change made?
Implement Arcadia's mapping for languages.

Unit

NA


app/services/language_builder.rb
```diff
@@ -5,17 +5,50 @@ class LanguageBuilder
...
+    end.uniq.compact.presence
+  end
+
+  private
+
+  attr_reader :languages
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
+  def iso_639?(lang)
```
jcoyne, line 33: 😱

jcoyne (approved):

app/services/language_builder.rb
justinlittman: And this is "normalized" data ...

PR #605, opened 2021-04-13
fix originInfo_place_placeTerm_tesim indexing.

### LINKED ISSUE #577: fix originInfo_place_placeTerm_tesim indexing
### What Needs to be Done

In `DescriptiveMetadataIndexer`, `originInfo_place_placeTerm_tesim` is not being populated correctly from cocina.

Finding the publicaion event in Cocina is expressed in PR #553;  retrieving the publication place will be more completely described by @arcadiafalcone here.
- Ensure we account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g.https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/origin_info_place_spec.rb. That is, if the code only expects "date.value" it won't find any values inside a structuredValue, a parallelValue (i.e. multi-lingual), and all that sort of stuff.

### User Impact

- we mistakenly stopped displaying the place of publication in Argo search result titles. Same for record views. And probably other places where we present citation info.
- We also use this field for searching.

justinlittman:
closes #577

## Why was this change made?
Mapping per Arcadia.

Unit

NA


app/services/event_place_builder.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+# Finds the place to index from publication events
+class EventPlaceBuilder
```
jcoyne, line 4: 👏

jcoyne (approved):

Issue #571, opened 2021-03-30
fix pub year indexing


ndushay:
### Background
There is a desire to keep certain DOR/Argo Solr index fields as similar to SearchWorks indexing as possible.  This was previously handled by using the same code in the `stanford-mods` gem for both this app and for SearchWorks to get the values for Solr fields.

PR #517 decoupled dor indexing from Fedora (MODS), but was done without this realization, so the algorithms for retrieving values for Solr from our (Cocina) descriptive metadata were inadvertently over-simplified.  Also, the original tests for the dor_indexing_app were not at all thorough in exploring different data conditions and how they would be indexed.

### What Needs to be Done
This ticket is to make the `sw_pub_date_facet_ssi` field contain values as much like the ones we were getting from the stanford-mods method as is reasonable.

Be certain PR #553 is merged before embarking on this ticket - it addresses part of this.

Per PR #517, `sw_pub_date_facet_ssi` Solr field was formerly populated by calling `pub_year_display_str` in stanford-mods (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/origin_info.rb#L47).  Tracing through the logic, you see it calls the `single_pub_year` method (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/origin_info.rb#L188).  This does fun processing first on the MODS dateIssued elements looking for a pub date and calling year_display_str (https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/origin_info.rb#L82) if there is a dateIssued element, then, if there isn't one, does similar logic on dateCreated MODS elements, and then looks at dateCaptured MODS elements.  PR #553 more or less deals with what would have been the dateIssued elements, but not the other flavors of dates.   Even for finding only what were the dateIssued elements in the MODS, there is a 3-tiered logic to identify these values in Cocina.  

- Work with Arcadia to find the corresponding information ^^ in cocina models.
    - ask her if there are nuances to the data
- Write tests that exercise the different use cases
- ensure you account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g.https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/origin_info_date_spec.rb.  That is, if the code only expects "date.value" it won't find any values inside a structuredValue, a parallelValue (i.e. multi-lingual), and all that sort of stuff.

### User Impact

[...]

PR #607, opened 2021-04-15
Spec to select preferred date from an event

arcadiafalcone:
## Why was this change made?
Part of #571

CircleCI

n/a


ndushay (changes requested):
Arcadia - can you make a separate file with your specs, and when I implement to get yours to pass, I will morph the two spec files together into the single one.

ndushay (approved):

PR #613, opened 2021-04-17
Implement prescribed mapping for pub date 

### LINKED ISSUE #571: fix pub year indexing

jcoyne:
## Why was this change made?
Fixes #571


spec/indexers/descriptive_metadata/pub_year_1e_spec.rb
```diff
@@ -365,8 +365,8 @@
         }
       end
 
-      xit 'selects first date' do
-        expect(doc).to include('sw_pub_date_facet_ssi' => '')
+      it 'selects first date' do
+        expect(doc).to include('sw_pub_date_facet_ssi' => '2020')
```
jcoyne, line 369: @arcadiafalcone is this the correct value?

spec/indexers/descriptive_metadata/pub_year_1e_spec.rb
```diff
@@ -570,8 +570,8 @@
         }
       end
 
-      xit 'selects date with status primary' do
-        expect(doc).to include('sw_pub_date_facet_ssi' => '2021-04-02')
+      it 'selects date with status primary' do
+        expect(doc).to include('sw_pub_date_facet_ssi' => '2022')
```
jcoyne, line 574: @arcadiafalcone this field is only an integer year, so I modified these few tests at the bottom.  Do they still look okay?

spec/indexers/descriptive_metadata/pub_year_1e_spec.rb
arcadiafalcone: Yes.

spec/indexers/descriptive_metadata/pub_year_1e_spec.rb
arcadiafalcone: Yes, these are all correct.

justinlittman (approved):

PR #618, opened 2021-04-22
Give spec file the proper filename so it is run by rspec

jcoyne:
## Why was this change made?

Previously the file did not end with `_spec.rb`


justinlittman (approved):
Nice.

PR #619, opened 2021-04-22
Implement prescribed mapping for publisher

### LINKED ISSUE #595: Update publisher indexing to select single value
Mapping should presume that value is pulled from single event. Multiple values within an event may be concatenated.

jcoyne:
## Why was this change made?
Fixes #595

### DIFF
--- a/spec/indexers/descriptive_metadata/event_contributor_publisher_spec.rb
+++ b/spec/indexers/descriptive_metadata/event_contributor_publisher_spec.rb
@@ -103,7 +103,7 @@ RSpec.describe DescriptiveMetadataIndexer do
         }
       end
 
-      xit 'selects publisher' do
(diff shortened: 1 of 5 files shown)


spec/indexers/descriptive_metadata/event_contributor_publisher_spec.rb
```diff
@@ -699,13 +699,13 @@
...
                         }
-                      ],
-                      role: [
-                        {
-                          value: 'publisher'
-                        }
                       ]
                     }
+                  ],
+                  role: [
```
jcoyne, line 705: The role attribute was in the wrong position. Arcadia confirmed that it should be a sibling of `name`.

justinlittman (approved):

PR #621, opened 2021-04-22
Subject topic

### LINKED ISSUE #578: fix topic_ssim and topic_tesim indexing
### What Needs to be Done

In `DescriptiveMetadataIndexer`, `topic_ssim` and `topic_tesim` are not being populated correctly from cocina.

These are the facet and search flavors of Solr fields for topic data.

Tagging @arcadiafalcone to get her input on where to find the topic data in Cocina;  Right now we are looking for `Cocina.description.subject` with `type` == `topic` ... but 
- ensure we account for the possibility of more complex data in the relevant cocina models by referring to the appropriate mapping specs in the dor-services-app, e.g. https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/mods/subject_topic_spec.rb. That is, if the code only expects "topic.value" it won't find any values inside a structuredValue, a parallelValue (i.e. multi-lingual), and all that sort of stuff.

### User Impact

- more complete data in topic facet and for searching

jcoyne:
## Why was this change made?

Fixes #578

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -188,7 +188,6 @@ class DescriptiveMetadataIndexer
   def nonstemmable_topics
     (
       TopicBuilder.build(Array(cocina.description.subject), filter: 'topic', remove_trailing_punctuation: true) +
-      # stemmable_topics + #.map { |topic| remove_trailing_punctuation(topic) } +
       TopicBuilder.build(Array(cocina.description.subject), filter: 'name')
     ).compact
   end
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -23,8 +23,8 @@ class DescriptiveMetadataIndexer
       'originInfo_date_created_tesim' => creation_date,
       'originInfo_publisher_tesim' => publisher_name,
       'originInfo_place_placeTerm_tesim' => event_place,
-      'topic_ssim' => topics,
-      'topic_tesim' => topics,
+      'topic_ssim' => nonstemmable_topics,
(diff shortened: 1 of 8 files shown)


app/indexers/descriptive_metadata_indexer.rb
```diff
@@ -181,8 +181,16 @@ def publisher_names_for(publication_or_event)
...
-  def topics
-    @topics ||= Array(cocina.description.subject).select { |subject| subject.type == 'topic' }.map(&:value).compact
+  def stemmable_topics
+    TopicBuilder.build(Array(cocina.description.subject), filter: 'topic')
+  end
+
+  def nonstemmable_topics
+    (
+      TopicBuilder.build(Array(cocina.description.subject), filter: 'topic', remove_trailing_punctuation: true) +
+      # stemmable_topics + #.map { |topic| remove_trailing_punctuation(topic) } +
```
justinlittman, line 191: cruft

app/services/title_builder.rb
```diff
@@ -1,20 +1,57 @@
 # frozen_string_literal: true
 
-class TitleBuilder
+class TitleBuilder # rubocop:disable Metrics/ClassLength
```
justinlittman, line 3: Slight preference for putting this in `rubocop.yml` to avoid rubocruft.

justinlittman (approved):
Approve pending consideration of comments.

app/services/title_builder.rb
jcoyne: Putting this in `.rubocop.yml` means it's something we want to keep exempt from the rules, but in this case we should fix it sometime, or acknowledge that it's a bit of a problem.

PR #627, opened 2021-04-29
Index access metadata from cocina

### LINKED ISSUE #510: Decouple RightsMetadataDatastreamIndexer from ActiveFedora
Depends on https://github.com/sul-dlss/argo/issues/2448

We must provide what DRA is doing now so we can have this facet:
Screen Shot 2021-03-12 at 5 12 29 PM

Ensure we can do this item: https://argo-stage.stanford.edu/view/druid:nn734fw1198
Screen Shot 2021-03-19 at 3 48 40 PM

```xml
<rightsMetadata>
  <access type="discover">
    <machine>
      <world/>
    </machine>
  </access>
  <access type="read">
    <machine>
      <group>stanford</group>
    </machine>
  </access>
  <access type="read">
    <machine>
      <world rule="no-download"/>
    </machine>
  </access>
  <access type="read">
    <file>nn734fw1198_md.pdf</file>
    <machine>
      <world/>
    </machine>
  </access>
  </rightsMetadata>
 ```

### PR DESCRIPTION
**Pull request by jcoyne**
## Why was this change made?

Fixes #510

### DIFF
--- a/app/indexers/rights_metadata_indexer.rb
+++ b/app/indexers/rights_metadata_indexer.rb
@@ -1,81 +1,23 @@
 # frozen_string_literal: true
 
 class RightsMetadataIndexer
-  attr_reader :resource, :cocina
+  attr_reader :cocina
 
-  def initialize(resource:, cocina:, **)
-    @resource = resource
+  def initialize(cocina:, **)
     @cocina = cocina
   end
 
   # @return [Hash] the partial solr document for rightsMetadata
-  # rubocop:disable Metrics/AbcSize
-  # rubocop:disable Metrics/CyclomaticComplexity
-  # rubocop:disable Metrics/MethodLength
-  # rubocop:disable Metrics/PerceivedComplexity
   def to_solr
     Rails.logger.debug "In #{self.class}"
 
-    solr_doc = {
-      'copyright_ssim' => resource.rightsMetadata.copyright,
-      'use_statement_ssim' => resource.rightsMetadata.use_statement
-    }
-
-    dra = resource.rightsMetadata.dra_object
-
-    solr_doc['rights_descriptions_ssim'] = [
-      dra.index_elements[:primary],
-
-      (dra.index_elements[:obj_locations_qualified] || []).map do |rights_info|
-        rule_suffix = rights_info[:rule] ? " (#{rights_info[:rule]})" : ''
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mjgiarlo** [comment]
I wonder if we should request that @justinlittman review this as the last of us to go deep into rights/access metadata. I too am happy to be roped in today/tomorrow if helpful!

File: app/indexers/rights_metadata_indexer.rb [hunk 1]
```diff
@@ -100,10 +42,34 @@ def to_solr
...
+                                             .map { |file| build_basic_access(file.access) }.uniq
+
+    [basic_access] + files.map { |result| "#{result} (file)" }
+  end
+
+  def build_basic_access(access)
+    basic_access = case access.access
+                   when 'citation-only'
+                     return 'citation'
+                   when 'dark'
```
justinlittman, line 59: Omit dark since will be covered by `else`?

app/indexers/rights_metadata_indexer.rb
```diff
@@ -100,10 +42,34 @@ def to_solr
     'http://opendatacommons.org/licenses/odbl/1.0/' => 'odc-odbl'
   }.freeze
 
+  def rights_descriptions
+    return 'controlled digital lending' if cocina.access.controlledDigitalLending
+
+    basic_access = build_basic_access(cocina.access)
+    files = Array(cocina.structural.contains).flat_map { |fs| fs.structural.contains }
```
justinlittman, line 49: `Array(fs.structural.contains)`?

justinlittman (approved):
Approve pending consideration of comments.

app/indexers/rights_metadata_indexer.rb
jcoyne: This would cause the function to return `dark (no-download)`, which is not the current value (although it is technically correct).

app/indexers/rights_metadata_indexer.rb
jcoyne: cast this to an array.  If it's `nil` it becomes an empty array.  If it's already an array return itself.

jcoyne: Ah, very good idea. Done.

app/indexers/rights_metadata_indexer.rb
justinlittman: Oh right, I see.

PR #635, opened 2021-04-30
More flexible mapping of marcgac and marc countries.

### LINKED ISSUE #633: [dor_indexing_app/prod] KeyError: key not found: "e-ru---"
Did you mean?  "e-ru"
## Backtrace

    line 54 of [PROJECT_ROOT]/app/services/geographic_builder.rb: place_from_code
    line 41 of [PROJECT_ROOT]/app/services/geographic_builder.rb: build_place
    line 34 of [PROJECT_ROOT]/app/services/geographic_builder.rb: block in build_place_nodes

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49898/faults/79112258)

justinlittman:
closes #633

## Why was this change made?
* Handle marcgac that include trailing spaces.
* Don't error when marcgac or country code not found.

Unit

NA

### DIFF
--- a/app/services/geographic_builder.rb
+++ b/app/services/geographic_builder.rb
@@ -48,12 +48,13 @@ class GeographicBuilder
   def place_from_code(node)
     return [] unless node.code && node.source
 
-    code = node.code.sub(/[^\w-]/, '') # remove any punctuation (except dash).
+    code_with_dashes = node.code.gsub(/[^\w-]/, '') # remove any punctuation (except dash).
+    code_without_dashes = code_with_dashes.sub(/-+$/, '') # remove trailing dashes.
     case node.source.code
     when 'marcgac'
-      [Marc::Vocab::GeographicArea.fetch(code)]
+      [Marc::Vocab::GeographicArea.fetch(code_with_dashes, nil) || Marc::Vocab::GeographicArea.fetch(code_without_dashes, nil)].compact
     when 'marccountry'
-      [Marc::Vocab::Country.fetch(code)]
+      [Marc::Vocab::Country.fetch(code_without_dashes, nil)].compact
     else
       []
     end
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ gem 'honeybadger', '~> 4.5'
 gem 'iso-639', '~> 0.3.5'
 # ISO-639-3
 gem 'language_list'
-gem 'marc-vocab', '~> 0.2.0'
+gem 'marc-vocab', '~> 0.3.0'
 gem 'okcomputer' # for monitoring
 gem 'puma', '~> 3.0' # app server
 gem 'rack-timeout', '~> 0.5.1'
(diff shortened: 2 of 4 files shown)


app/services/geographic_builder.rb
```diff
@@ -48,12 +48,13 @@ def build_place(node)
   def place_from_code(node)
     return [] unless node.code && node.source
 
-    code = node.code.sub(/[^\w-]/, '') # remove any punctuation (except dash).
+    code_with_dashes = node.code.gsub(/[^\w-]/, '') # remove any punctuation (except dash).
+    code_without_dashes = code_with_dashes.sub(/-+$/, '') # remove trailing dashes.
     case node.source.code
     when 'marcgac'
-      [Marc::Vocab::GeographicArea.fetch(code)]
+      [Marc::Vocab::GeographicArea::TERMS[code_with_dashes] || Marc::Vocab::GeographicArea::TERMS[code_without_dashes]].compact
```
jcoyne, line 55: Please use `fetch` with a default (`nil`) here rather than addressing the constants (which is an implementation detail).

app/services/geographic_builder.rb
```diff
@@ -48,12 +48,13 @@ def build_place(node)
   def place_from_code(node)
     return [] unless node.code && node.source
 
-    code = node.code.sub(/[^\w-]/, '') # remove any punctuation (except dash).
+    code_with_dashes = node.code.gsub(/[^\w-]/, '') # remove any punctuation (except dash).
```
jcoyne, line 51: Should this be in DSA?  Shouldn't the cocina model only have valid codes?

app/services/geographic_builder.rb
```diff
@@ -48,12 +48,13 @@ def build_place(node)
...
+    code_without_dashes = code_with_dashes.sub(/-+$/, '') # remove trailing dashes.
```
jcoyne, line 52: Should this be in DSA? Shouldn't the cocina model only have valid codes?

app/services/geographic_builder.rb
justinlittman: Duh.

justinlittman: Oh wait, it doesn't take a default.

app/services/geographic_builder.rb
justinlittman: Depends on how you define "valid" I guess ...

I'm not opposed to a separate ticket in dor-services-app to consider.

app/services/geographic_builder.rb
jcoyne: `bundle update marc-vocab`

app/services/geographic_builder.rb
jcoyne: I would expect that valid means any marcgac codes to be on the list that LOC publishes.

app/services/geographic_builder.rb
jcoyne: The 0.3.0 version does.

app/services/geographic_builder.rb
justinlittman: https://github.com/sul-dlss/dor-services-app/issues/2805

jcoyne (approved):

PR #639, opened 2021-05-01
[TEMP, don't merge] temporarily silence honeybadger alert so we don't use quota so fast on stage

jmartin-sul:
## Why was this change made?

lots of alerts for unrecognized place code on stage.  since it's just stage, we can take a look monday, without using up lots of our alert quota

https://app.honeybadger.io/projects/49898/faults/79112314
https://app.honeybadger.io/projects/49898/faults/79122054

i just deployed it to stage and the alerts seem to have stopped 🤞 ...

n/a

### DIFF
--- a/app/services/geographic_builder.rb
+++ b/app/services/geographic_builder.rb
@@ -58,7 +58,9 @@ class GeographicBuilder
       []
     end
   rescue KeyError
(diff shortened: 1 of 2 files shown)


jcoyne:
This was occurring because DSA hasn't been redeployed on stage which takes care of this bad data.

jcoyne:
Can we delete this now?

PR #645, opened 2021-05-13
Only index the latest release

jcoyne:
## Why was this change made?
Fixes https://github.com/sul-dlss/argo/issues/2748

### DIFF
--- a/app/indexers/releasable_indexer.rb
+++ b/app/indexers/releasable_indexer.rb
@@ -10,20 +10,19 @@ class ReleasableIndexer
   end
 
   # @return [Hash] the partial solr document for releasable concerns
+  # TODO: sort of worried about the performance impact in bulk reindex
+  # situations, since released_for recurses all parent collections.  jmartin 2015-07-14
   def to_solr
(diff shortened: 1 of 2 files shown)


peetucket (dismissed):

peetucket:
Testing on -stage right now and just noticed this object went from a double Seachworks and Earthworks to just Earthworks: https://argo-stage.stanford.edu/view/druid:bg512yv8189

Cocina model shows both targets have release tags.  I'll investigate

peetucket:
oh, i see, that object works as expected...the last release to searchworks was "false", so it's no longer released there

peetucket:
so it should show any targets, only once, and only if still released

and in fact this object in stage used to show "Searchworks and Searchworks" and now shows nothing, which is correct, because the latest tag now shows it was eventually "unreleased" from searchworks https://argo-stage.stanford.edu/view/druid:bb234xx1459

LGTM - pending confirmation of expected behavior from @andrewjbtw

peetucket (approved):

PR #659, opened 2021-06-25
Protect against versionMetadata not being found

jcoyne:
## Why was this change made?
This is a rare occurrence that only occurs on non-cocina compliant objects when they also don't have a versionMetadata datastream.

see https://app.honeybadger.io/projects/49898/faults/79971822


ndushay:
build failed - rubocop

PR #688, opened 2021-10-07
Index releases from the collection onto the items

### LINKED ISSUE #686: Items released to SW via collection release are not being indexed correctly in Argo
**Describe the bug**

Items that are released to SW via a collection release are not being indexed properly in Argo. For example:

This item has been released to SW: https://searchworks.stanford.edu/view/dd331tq0265
But Argo does not show that it has been released: https://argo.stanford.edu/view/druid:dd331tq0265

It was released automatically because the collection that contains it has been released under the setting that releases all members of that collection: https://argo.stanford.edu/view/druid:bc300zz4971

If you look at the members of the collection in Argo as a search, you don't see the "Released to" facet showing that they have been released:
https://argo.stanford.edu/catalog?f%5Bis_member_of_collection_ssim%5D%5B%5D=info%3Afedora%2Fdruid%3Abc300zz4971

But the SW collection page shows that all 6 items have been released: https://searchworks.stanford.edu/view/bc300zz4971

**Background**

There are two ways to release an item to SearchWorks/Earthworks:

- Release the item itself
- Release all items in the collection to which the item belongs

In both cases, Argo should report that the item has been released to SW. This means that:

- On the Argo show page, the "Released to" value is filled in with "SearchWorks" or "EarthWorks" or both
- On the Argo show page, the list of links in the top left corner should include a link to the SW record
- In the Argo facets, the item should appear in the "Released to" facet when it comes up in a search

Where these two release methods differ is in how the release data is stored:

1. For specific item releases
 - a "release" tag is added to the identityMetadata (mapped to `.identification` in cocina): 
```xml
<release what="self" to="Searchworks" who="amyhodge" when="2020-02-26T16:40:16Z">true</release>
```
 - the public XML published to the Purl includes an xml block for `releaseData`
```xml
<releaseData>
<release to="Searchworks">true</release>
</releaseData>
```

2. For items released via a collection release

- only the public XML is updated to include the `releaseData` tags (as above). I believe this happens during the "release-members" step in the releaseWF on the collection object (no releaseWF is run on the item itself)
- no change is made to the `identityMetadata` or Cocina data

**User Impact**
Argo users cannot reliably tell if an item has been released to SW. This can lead to:

- unnecessary, manually-run item releases (if a user doesn't realize something was already released)
- difficulties in determining when a release failed (vs. when it succeeded but wasn't indexed)
- inaccurate reporting of numbers of released items (SW release is a key metric that the library tracks)

**To Reproduce**

1. See above for an example of an item released via collection that isn't indexed properly. We don't have Argo stage connected to SW stage, so we can't test this without actually releasing items in production.

**Expected behavior**
Items that have been released to SW should be indexed in Argo to show that they've been released.

**Additional context**
I suspect what is happening here is that the Argo indexer is only indexing releases that appear in the Cocina and not indexing releases where only the Purl was updated. I wonder if the previous Fedora-based indexing looked at both the Purl and the identityMetadata.

The release process itself seems to be working, based on a spot check of recently-released items. Items are still making it to SW and EW when released via collection.

jcoyne:
This was approached, by having the Indexer load any collections upfront and passing the loaded collections to the indexers.  Then these collections could have different parts indexed, without having to load them more than one time.

## Why was this change made?

Fixes #686 

Andrew is reviewing on stage.

### DIFF
--- a/app/indexers/releasable_indexer.rb
+++ b/app/indexers/releasable_indexer.rb
@@ -11,7 +11,10 @@ class ReleasableIndexer
   # @return [Hash] the partial solr document for releasable concerns
   def to_solr
     Rails.logger.debug "In #{self.class}"
-    values = tags_from_item + tags_from_collection
+
+    # Item tags have precidence over collection tags, so if the collection is release=true
+    # and the item is release=false, then it is not released
+    values = tags_from_collection.merge(tags_from_item).values.select(&:release).map(&:to)
 
     return {} if values.blank?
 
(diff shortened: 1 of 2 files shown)


jcoyne:
I don't think this change requires any new integration test as the change is confined entirely to this code base.

PR #706, opened 2022-01-06
Updating to ruby3

edsu:
## Why was this change made?

ruby3.0 techstack upgrades

Ran tests locally under ruby v2.7.5 and ruby v3.0.3. The build passed in [CircleCI](https://app.circleci.com/pipelines/github/sul-dlss/dor_indexing_app/386/workflows/bb103167-a4a3-463b-98a2-7b1ef0beafc2/jobs/623).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,14 +4,14 @@ executors:
     environment:
       IMAGE_NAME: suldlss/dor-indexing-app
     docker:
-    - image: circleci/buildpack-deps:stretch
+    - image: cimg/base
 orbs:
   ruby: circleci/ruby@0.1.2
 
 jobs:
   test:
     docker:
-      - image: circleci/ruby:3.0.3-node
+      - image: cimg/ruby:3.0.3-node
     executor: ruby/default
     steps:
       - checkout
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -11,3 +11,7 @@
 config/certs
 config/settings/*.local.yml
 /spec/examples.txt
+
+
+.rbenv-gemsets
```
jcoyne, line 16: would this go better in a global .gitignore? https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

edsu: Is there a standard .gitignore we all use?

.circleci/config.yml
```diff
@@ -4,14 +4,14 @@ executors:
     environment:
       IMAGE_NAME: suldlss/dor-indexing-app
     docker:
-    - image: circleci/buildpack-deps:stretch
+    - image: cimg/base
```
ndushay, line 7: not sure this is going to work when this PR is merged to main and a docker image needs to be created;  see https://github.com/sul-dlss/dlme/pull/1382 and maybe talk to @aaron-collier

jcoyne (approved):

PR #722, opened 2022-01-26
index created at date

### LINKED ISSUE #719: Index date created
Similar to how we have `Last-Modifed` header here.
https://github.com/sul-dlss/dor_indexing_app/blob/ed395eacff8d1108135069eaa339f84204be0200/app/indexers/data_indexer.rb#L8

We will create a new header called `X-Created-At` that will support this value.

This will be used to show the Created date in Argo because not all objects (apo's, collections, agreements) have registration dates.

peetucket:
## Why was this change made?

Fixes #719 

See https://github.com/sul-dlss/dor-services-app/pull/3198/files for where this date comes from

Question: I just made up a name of the solr field to index to called `created_at_dttsi`  - I presume this is a new solr field that will be shown in argo, so the name doesn't have to match anything that currently exists.

Updated existing tests


mjgiarlo (approved):

mjgiarlo:
Holding off merging since @justinlittman requested a review from @jcoyne

jcoyne (approved):

jcoyne (approved):

Issue #729, opened 2022-02-02
Reimplement reindexer script


jcoyne:
Consider using `eye` as the daemon for this. I believe the access team uses this for https://github.com/sul-dlss/searchworks_traject_indexer

This should reimplement this: https://github.com/sul-dlss/dor-camel-routes/blob/6c0502d523bd2cc4b7ca588c1149cccc3b34dd4d/deploy/dor_prod_background_reindexing.xml#L11

jcoyne:
Fixed by #730

PR #730, opened 2022-02-02
Rolling indexer

jcoyne:
## Why was this change made?
In the past we had a camel route to reindex the old data. This allows us to migrate away from that.

Ref #729


app/controllers/dor_controller.rb
```diff
@@ -24,40 +23,7 @@ def queue_size
 
   private
 
-  # retrieves a single Dor object by pid, indexes the object to solr, does some logging
-  # doesn't commit automatically.
-  def reindex_pid(pid, add_attributes:)
```
jcoyne, line 29: This has all been extracted to Indexer.  The old Indexer has been renamed DocumentBuilder

app/services/document_builder.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
```
jcoyne, line 1: This is just moved from Indexer and renamed to DocumentBuilder. No new code was added.

app/services/indexer.rb
```diff
@@ -1,79 +1,48 @@
 # frozen_string_literal: true
 
-require 'dry/monads/maybe'
```
jcoyne, line 3: This code moved to DocumentBuilder.

app/services/indexer.rb
```diff
@@ -1,79 +1,48 @@
...
-  def self.for(model:, metadata:)
-    Rails.logger.debug { "Fetching indexer for #{model.type}" }
-    parent_collections = load_parent_collections(model)
-    INDEXERS.fetch(model.type, ITEM_INDEXER).new(id: model.externalIdentifier,
-                                                 cocina: model,
-                                                 parent_collections: parent_collections,
-                                                 metadata: metadata)
+  # retrieves a single Dor object by pid, indexes the object to solr, does some logging
+  # doesn't commit automatically.
+  def reindex_pid(pid, add_attributes:)
```
jcoyne, line 12: This code moved from DorController.

bin/rolling_index
```diff
@@ -0,0 +1,20 @@
...
+    indexer = Indexer.new(solr: solr_conn)
+
+    response = solr_conn.get 'select', params: QUERY
+    response['response']['docs'].each do |doc|
+      puts "Doc #{doc}"
+      indexer.reindex_pid(doc['id'], add_attributes: { commitWithin: 1000 })
+    end
+    sleep(5)
+  end
+end
```
mjgiarlo, line 20: Hi! This looks solid to me. I'm curious about reaching for `daemons` rather than, say, a scheduled task. Can you say more about that?

jcoyne: This is constantly running, so it's a daemon. It's not scheduled.

mjgiarlo: @jcoyne A benefit of the daemon over the scheduled task is lower/no risk of one task running and not completing before the next task is invoked. A cost is needing to make sure this process is run at start time and boot time and restarted at deploy time. Do we need any sort of process management on top to help manage this?

jcoyne: Yeah, I think we'll need to work with ops to get a systemd(?) script to run this.

mjgiarlo: @jcoyne and we'll prolly also need some capistrano work to use said systemd script. Would you mind ticketing if not already done? That assuages my concern. Approving.

jcoyne: Sure I will do that.

mjgiarlo (approved):

PR #731, opened 2022-02-03
Add jobs for reindexing from RabbitMQ messages

### LINKED ISSUE #728: Trigger indexing when a RabbitMQ message is created
Currently indexing depends on Fedora 3 sending ActiveMQ messages, which are routed by Karaf to dor_indexing_app.  We will replace this functionality by enabling dor_indexing_app to listen to RabbitMQ messages. (Similar to how h2 does) and turn off the Karaf route that sends ActiveMQ messages to dor_indexing_app

DSA already sends messages: https://github.com/sul-dlss/dor-services-app/blob/e8d7595f96c62b8b4a7c13a98618c2c66a521504/app/services/notifications/object_updated.rb

Workflow-server-rails already sends messages https://github.com/sul-dlss/workflow-server-rails/blob/e328f768446aa37dcd067d830f9833c7c4f36c57/app/services/send_update_message.rb#L9

jcoyne:
## Why was this change made?
Fixes #728

### DIFF
--- a/app/jobs/reindex_by_druid_job.rb
+++ b/app/jobs/reindex_by_druid_job.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+# Reindexes an object given a druid
+class ReindexByDruidJob
+  include Sneakers::Worker
+  # This worker will connect to "dor.indexing-by-druid" queue
+  # env is set to nil since by default the actual queue name would be
+  # "dor.indexing-by-druid_development"
+  from_queue 'dor.indexing-by-druid', env: nil
+
+  def work(msg)
+    druid = druid_from_message(msg)
+    # Since we don't have the metadata (namely created_at) in the message,
+    # we need another API call. :(
+    indexer = Indexer.new(solr: solr)
+    indexer.reindex_pid druid, add_attributes: { commitWithin: 1000 }
+    ack!
+  end
+
+  def solr
+    RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solrizer_url)
+  end
+
+  def druid_from_message(str)
+    JSON.parse(str).fetch('druid')
+  end
+end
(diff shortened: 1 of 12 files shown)


app/jobs/reindex_by_druid_job.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# Reindexes an object given a druid
+class ReindexByDruidJob
+  include Sneakers::Worker
+  # This worker will connect to "dor.indexing-with-model" queue
+  # env is set to nil since by default the actual queue name would be
+  # "dor.indexing-with-model_development"
+  from_queue 'dor.indexing-by-druid', env: nil
```
mjgiarlo, line 9: Should this comment block have `s/with-model/by-druid/g`?

config/settings.yml
```diff
@@ -20,3 +20,10 @@ workflow_url: 'https://workflow.example.edu/'
 dor_services:
   url: 'http://localhost:3003'
   token: 'not-a-real-token'
+
+
+rabbitmq:
+  hostname: localhost
+  vhost: /
+  username: guest
+  password: guest
```
mjgiarlo, line 29: Looks good. I see that QA and Stage in shared_configs are configured slightly differently. Looks like one has a vhost and one doesn't. Intended? And I assume doing this for prod is in your plan later on.

mjgiarlo (approved):
I have a minor suggestion re: a code comment but am approving

config/settings.yml
jcoyne: It's a PR awaiting review: https://github.com/sul-dlss/shared_configs/pull/1723

app/jobs/reindex_by_druid_job.rb
jcoyne: Yep, I've made this fix.

PR #746, opened 2022-02-11
Index constituent relationships

### LINKED ISSUE #743: Index constituents of a virtual object
When a non-empty structural > hasMemberOrders[0] > members, the members (druids) should be indexed as `has_constituents_ssim`.

This will be used so DSA can use to generate public XML whereas a constituent can find the virtual object it is part of.

Add an entry documenting this to the Google doc linked in the README.

jcoyne:
## Why was this change made?

Fixes #743

### DIFF
--- a/app/indexers/data_indexer.rb
+++ b/app/indexers/data_indexer.rb
@@ -23,6 +23,9 @@ class DataIndexer
       solr_doc['has_model_ssim'] = legacy_model
       solr_doc['is_governed_by_ssim'] = legacy_apo
       solr_doc['is_member_of_collection_ssim'] = legacy_collections
+
+      # Used so that DSA can generate public XML whereas a constituent can find the virtual object it is part of.
+      solr_doc['has_constituents_ssim'] = virtual_object_constituents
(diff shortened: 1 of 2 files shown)


app/indexers/data_indexer.rb
```diff
@@ -43,6 +47,12 @@ def legacy_collections
     end
   end
 
+  def virtual_object_constituents
+    return unless cocina.dro?
+
+    cocina.structural.hasMemberOrders.first&.members
```
mjgiarlo, line 53: Can we be sure every DROStructural has a `hasMemberOrders`? Wondering if we need more safe navigation on this line. 🤷🏻‍♂️

mjgiarlo (approved):
have a question about safe navigation but otherwise 👍🏻

app/indexers/data_indexer.rb
jcoyne: I'm pretty sure the tests cover this case. Every DROStructural will have hasMemberOrders and it will return an array: https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/dro_structural.rb#L7

mjgiarlo: Sounds good.

PR #748, opened 2022-02-14
[SPIKE] Index virtual objects as both items and virtual objects

mjgiarlo:
## Why was this change made?

This would make it easier for Argo users to find and operate in bulk on virtual objects. Whether it is valuable, or has other risks or side effects, would be worth discussing. I could also be convinced that we could/should use the existing has_constituents_ssim field for this somehow?

CI

None

### DIFF
--- a/app/indexers/identity_metadata_indexer.rb
+++ b/app/indexers/identity_metadata_indexer.rb
@@ -9,10 +9,10 @@ class IdentityMetadataIndexer
 
   # @return [Hash] the partial solr document for identityMetadata
   def to_solr
-    return { 'objectType_ssim' => [object_type] } if object_type == 'adminPolicy' || cocina_object.identification.nil?
+    return { 'objectType_ssim' => object_types } if object_types.include?('adminPolicy') || cocina_object.identification.nil?
 
     {
-      'objectType_ssim' => [object_type],
+      'objectType_ssim' => object_types,
       'dor_id_tesim' => [source_id_value, barcode, catkey].compact,
       'identifier_ssim' => prefixed_identifiers,
       'identifier_tesim' => prefixed_identifiers,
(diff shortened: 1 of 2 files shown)


app/indexers/identity_metadata_indexer.rb
```diff
@@ -33,21 +33,24 @@ def source_id_value
...
     when Cocina::Models::AdminPolicy
-      'adminPolicy'
+      ['adminPolicy']
     when Cocina::Models::Collection
-      'collection'
+      ['collection']
     else
-      cocina_object.type == Cocina::Models::Vocab.agreement ? 'agreement' : 'item'
+      return ['agreement'] if cocina_object.type == Cocina::Models::Vocab.agreement
+      return ['item'] unless cocina_object.structural&.hasMemberOrders&.first&.members
```
justinlittman, line 51: `.present?` to avoid `[]`?

justinlittman (approved):
Approve pending consideration of comment and would like @jcoyne to weigh in whether this might have other effects.

jcoyne:
Is there a reason we can't use the existing has_constituents_ssim for this?

app/indexers/identity_metadata_indexer.rb
```diff
@@ -9,10 +9,10 @@ def initialize(cocina:, **)
 
   # @return [Hash] the partial solr document for identityMetadata
   def to_solr
-    return { 'objectType_ssim' => [object_type] } if object_type == 'adminPolicy' || cocina_object.identification.nil?
+    return { 'objectType_ssim' => object_types } if object_types.include?('adminPolicy') || cocina_object.identification.nil?
 
     {
-      'objectType_ssim' => [object_type],
+      'objectType_ssim' => object_types,
```
jcoyne, line 15: This field is used by Argo to determine what partials to display and I'd be a bit hesitant to add values to it that are not expected.

jcoyne (changes requested):

mjgiarlo:
@jcoyne 💬 
> Is there a reason we can't use the existing has_constituents_ssim for this?

That is a question I had as well up in the PR description. I can look into it.

mjgiarlo:
Superseded by https://github.com/sul-dlss/argo/pull/3110, closing

PR #753, opened 2022-02-16
Set up RabbitMQ queues as part of deployment

mjgiarlo:

This prevents us having to do this manually on every node in every environment. Also, move common capistrano config out of env-specific configs.

CI, tested in QA

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
 #   https://github.com/capistrano/rbenv
 #   https://github.com/capistrano/chruby
 #   https://github.com/capistrano/bundler
-#   https://github.com/capistrano/rails
 #
 # require 'capistrano/rvm'
 # require 'capistrano/rbenv'
 # require 'capistrano/chruby'
-# require 'capistrano/rails/assets'
-# require 'capistrano/rails/migrations'
 
 require 'capistrano/bundler'
 require 'capistrano/honeybadger'
-# most rails apps need this, but this one doesn't because there's no db to
-# migrate, and there are no assets to precompile
-# (https://github.com/capistrano/rails#usage)
-#
-# require 'capistrano/rails'
 require 'capistrano/passenger'
 require 'capistrano/shared_configs'
 require 'dlss/capistrano'
--- a/README.md
+++ b/README.md
@@ -25,13 +25,15 @@ RAILS_ENV=production bin/rolling_index stop
 
 See https://sul-dlss.github.io/dor_indexing_app/
 
-## Setup RabbitMQ
-You must set up the durable rabbitmq queues that bind to the exchange where workflow messages are published.
+## RabbitMQ Setup
+
+NOTE: This is done automatically by Capistrano on deployment, so one generally shouldn't need to run this task manually. This is for informational purposes only.
+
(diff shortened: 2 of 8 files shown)


Capfile
```diff
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
 #   https://github.com/capistrano/rbenv
 #   https://github.com/capistrano/chruby
 #   https://github.com/capistrano/bundler
-#   https://github.com/capistrano/rails
```
mjgiarlo, line 19: This app does not use capistrano-rails so I'm not convinced we need all this cruft hanging around. If we ever add DB migrations or assets to DIA, we can always add this back.

Capfile
```diff
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
...
 # require 'capistrano/rbenv'
 # require 'capistrano/chruby'
-# require 'capistrano/rails/assets'
-# require 'capistrano/rails/migrations'
 
 require 'capistrano/bundler'
 require 'capistrano/honeybadger'
-# most rails apps need this, but this one doesn't because there's no db to
-# migrate, and there are no assets to precompile
-# (https://github.com/capistrano/rails#usage)
```
mjgiarlo, line 31: <iframe src="https://giphy.com/embed/igR5863TALcSk" width="480" height="383" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/dancing-shark-forum-igR5863TALcSk">via GIPHY</a></p>

README.md
```diff
@@ -25,13 +25,15 @@ RAILS_ENV=production bin/rolling_index stop
 
 See https://sul-dlss.github.io/dor_indexing_app/
 
-## Setup RabbitMQ
-You must set up the durable rabbitmq queues that bind to the exchange where workflow messages are published.
+## RabbitMQ Setup
+
+NOTE: This is done automatically by Capistrano on deployment, so one generally shouldn't need to run this task manually. This is for informational purposes only.
```
jcoyne, line 30: I don't think we want to do this on every deploy.  We only want to run this when the bindings change.

jcoyne (changes requested):

Capfile
```diff
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
...
-#
-# require 'capistrano/rails'
```
mjgiarlo, line 33: let it go

README.md
mjgiarlo: @jcoyne Can you say why we might not want to do this every time? It seems pretty cheap. Maybe it has side effects we don't want?

mjgiarlo: @jcoyne I'm going to noodle on this a bit and see if I can't get a quick improvement in. Stay tuned.

mjgiarlo: @jcoyne meh, not sure it's worth any more noodling time. I'm still interested in your answer to what I asked above, tho.

jcoyne: Deploying is something we do several times a week, but I don't anticipate we need to add any further bindings.  Each step in the deploy adds a few seconds and an extra point where problems can arise. I think having small special purpose tools makes things work better and easier to understand.

PR #754, opened 2022-02-17
index rights for APOs

peetucket:

Part of https://github.com/sul-dlss/argo/issues/3121

- Index "Dark" into the rights description field for all APOs to fix the display of object rights for APOs
- Derive the display of default object rights for display on APOs into a new solr field

Will require updating Argo to recognize this new field and use it to add back the display of default object rights on APOs that was removed in https://github.com/sul-dlss/argo/pull/3123 

- Existing tests for items/collections (indexing should unchanged despite the refactor)
- Update tests for APO indexing which has new fields

### DIFF
--- a/app/indexers/default_object_rights_indexer.rb
+++ b/app/indexers/default_object_rights_indexer.rb
@@ -13,7 +13,9 @@ class DefaultObjectRightsIndexer
 
     {
       'use_statement_ssim' => use_statement,
-      'copyright_ssim' => copyright
+      'copyright_ssim' => copyright,
+      'rights_descriptions_ssim' => 'Dark',
+      'default_rights_descriptions_ssim' => RightsDescriptionBuilder.build(cocina)
     }
   end
 
--- a/app/services/collection_rights_description_builder.rb
+++ b/app/services/collection_rights_description_builder.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# Rights description builder for collections
+class CollectionRightsDescriptionBuilder
+  def self.build(cocina)
+    new(cocina).build
+  end
+
+  def initialize(cocina)
+    @cocina = cocina
+  end
+
+  def build
+    case cocina.access.access
+    when 'world'
+      'world'
+    else
+      'dark'
+    end
+  end
+
+  private
+
+  attr_reader :cocina
+end
--- a/app/services/rights_description_builder.rb
+++ b/app/services/rights_description_builder.rb
@@ -7,18 +7,19 @@ class RightsDescriptionBuilder
 
   def initialize(cocina)
     @cocina = cocina
+    @root_access_node = cocina.admin_policy? ? cocina.administrative.defaultAccess : cocina.access
   end
 
   def build
-    cocina.dro? ? rights_descriptions_for_item : rights_descriptions_for_collection
+    cocina.collection? ? rights_descriptions_for_collection : rights_descriptions_for_item_and_apo
   end
 
   private
 
-  attr_reader :cocina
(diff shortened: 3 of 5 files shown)


app/indexers/default_object_rights_indexer.rb
```diff
@@ -13,7 +13,9 @@ def to_solr
 
     {
       'use_statement_ssim' => use_statement,
-      'copyright_ssim' => copyright
+      'copyright_ssim' => copyright,
+      'rights_descriptions_ssim' => 'Dark',
+      'default_rights_descriptions_ssim' => RightsDescriptionBuilder.build(cocina)
     }
```
peetucket, line 19: APOs are always "Dark" for their own rights, and get a new field for user display of default object rights

app/indexers/default_object_rights_indexer.rb
```diff
@@ -13,7 +13,9 @@ def to_solr
 
     {
       'use_statement_ssim' => use_statement,
-      'copyright_ssim' => copyright
+      'copyright_ssim' => copyright,
+      'rights_descriptions_ssim' => 'Dark',
+      'default_rights_descriptions_ssim' => RightsDescriptionBuilder.build(cocina)
```
jcoyne, line 18: Shouldn't the parameter be `cocina.defaultAccess` ?

app/services/rights_description_builder.rb
```diff
@@ -7,36 +7,39 @@ def self.build(cocina)
 
   def initialize(cocina)
```
jcoyne, line 8: I think this would be a lot easier to understand if this was separated into two classes, one for collections and one for DROAcess or AdminPolicyDefaultAccess. That would make for many fewer conditionals.

jcoyne (changes requested):

app/services/rights_description_builder.rb
peetucket: Ok, collections in one class, items/APOs in the other?  Items/APOs are more alike.

app/indexers/default_object_rights_indexer.rb
peetucket: No because items need to dig into places other than the access node (they go into structural), which is why I left the paramter as the root of the cocina model and then deciding where to find the rights based on the type (APO vs item).

see for example: https://github.com/sul-dlss/dor_indexing_app/blob/index-rights-description-for-apos/app/services/rights_description_builder.rb#L77-L82

app/services/rights_description_builder.rb
jcoyne: Yes, exactly.

spec/indexers/default_object_rights_indexer_spec.rb
```diff
@@ -34,6 +37,8 @@
       expect(doc).to match a_hash_including('use_statement_ssim' =>
         'Rights are owned by Stanford University Libraries.')
       expect(doc).to match a_hash_including('copyright_ssim' => 'Additional copyright info')
+      expect(doc).to match a_hash_including('rights_descriptions_ssim' => 'Dark')
+      expect(doc).to match a_hash_including('default_rights_descriptions_ssim' => ['location: spec'])
     end
```
peetucket, line 42: additional fields now indexed into APOs

app/services/collection_rights_description_builder.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Rights description builder for collections
+class CollectionRightsDescriptionBuilder
```
jcoyne, line 4: 👍

jcoyne (approved):

PR #757, opened 2022-02-18
Remove capistrano-rails and move common config to common config

mjgiarlo:

This app does not use capistrano-rails so I'm not convinced we need all this cruft hanging around. If we ever add DB migrations or assets to DIA, we can always add this back. In the meantime, let it go.

Also, move common Capistrano config to... common Capistrano config. Having the Rails env and bundler without args in specific env configs is a relic of past practice where we had custom Rails envs.

CI

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
 #   https://github.com/capistrano/rbenv
 #   https://github.com/capistrano/chruby
 #   https://github.com/capistrano/bundler
-#   https://github.com/capistrano/rails
 #
 # require 'capistrano/rvm'
 # require 'capistrano/rbenv'
 # require 'capistrano/chruby'
-# require 'capistrano/rails/assets'
-# require 'capistrano/rails/migrations'
 
 require 'capistrano/bundler'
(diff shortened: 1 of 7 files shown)


Capfile
```diff
@@ -16,21 +16,13 @@ install_plugin Capistrano::SCM::Git
...
-# require 'capistrano/rails/assets'
-# require 'capistrano/rails/migrations'
 
 require 'capistrano/bundler'
 require 'capistrano/honeybadger'
-# most rails apps need this, but this one doesn't because there's no db to
-# migrate, and there are no assets to precompile
-# (https://github.com/capistrano/rails#usage)
-#
-# require 'capistrano/rails'
```
ndushay, line 33: I would have left this stuff in (except L29-xx), but whatevs.

Gemfile
```diff
@@ -54,7 +54,6 @@ group :deployment do
   gem 'capistrano', '~> 3.0'
   gem 'capistrano-bundler'
   gem 'capistrano-passenger'
-  gem 'capistrano-rails'
```
ndushay, line 57: yay!

ndushay (approved):

PR #769, opened 2022-03-02
[HOLD for integration test] DorController#reindex: allow caller to provide cocina

### LINKED ISSUE #765: Allow a HTTP reindexing request to be made where the caller provides COCINA
This will allow DSA to provide the document that needs to be indexed, rather than have DSA call DIA which calls back to DSA.

jmartin-sul:

fix #765 

- [x] unit tests
- [ ] infra integration tests

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -2,17 +2,21 @@
 
 # Main controller of application
 class DorController < ApplicationController
+  include Dry::Monads[:result]
+
+  class CocinaModelBuildError < StandardError; end
+
   def reindex
-    indexer = Indexer.new(solr: solr, pid: params[:pid])
-    cocina_with_metadata = indexer.fetch_model_with_metadata
-    @solr_doc = indexer.reindex_pid(
-      add_attributes: { commitWithin: params.fetch(:commitWithin, 1000).to_i },
-      cocina_with_metadata: cocina_with_metadata
-    )
-    indexer.commit unless params[:commitWithin] # reindex_pid doesn't commit, but callers of this method may expect the update to be committed immediately
+    @indexer = Indexer.new(solr: solr, pid: params[:pid])
+
+    reindex_pid
     render status: :ok, plain: "Successfully updated index for #{params[:pid]}"
   rescue Dor::Services::Client::NotFoundResponse, Rubydora::RecordNotFound
     render status: :not_found, plain: 'Object does not exist in the repository'
+  rescue CocinaModelBuildError => e
+    request.session # TODO: calling this as a hack to address bad Rails/HB interaction, remove when https://github.com/rails/rails/issues/43922 is fixed
+    Honeybadger.notify('Error building Cocina model', context: { druid: params[:pid], build_error: e.cause.message }, backtrace: e.cause.backtrace)
+    render status: :unprocessable_entity, plain: "Error building Cocina model for #{params[:pid]}"
   end
 
   def delete_from_index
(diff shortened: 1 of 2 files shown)


app/controllers/dor_controller.rb
```diff
@@ -2,17 +2,21 @@
...
-    )
-    indexer.commit unless params[:commitWithin] # reindex_pid doesn't commit, but callers of this method may expect the update to be committed immediately
+    @indexer = Indexer.new(solr: solr, pid: params[:pid])
+
+    reindex_pid
     render status: :ok, plain: "Successfully updated index for #{params[:pid]}"
   rescue Dor::Services::Client::NotFoundResponse, Rubydora::RecordNotFound
     render status: :not_found, plain: 'Object does not exist in the repository'
+  rescue CocinaModelBuildError => e
+    request.session # TODO: calling this as a hack to address bad Rails/HB interaction, remove when https://github.com/rails/rails/issues/43922 is fixed
```
jmartin-sul, line 17: so i lucked out (?) and found this problem because i wrote my build error test before doing an `allow`/`expect` on the HB notification, and i was perplexed by the fact that i was getting `NoMethodError: undefined method `enabled?' for {}:Hash` on the line below.  after confirming that i was indeed calling `Honeybadger.notify` the same way as i saw it called in our other apps, and after trying (unsuccessfully) to reproduce this weird behavior by getting rid of a similar `allow(Honeybadger).to receive(:notify)` in `spec/services/document_builder_spec.rb` (which exercises `app/indexers/identifiable_indexer.rb`), i did some web searching, and came across https://github.com/rails/rails/issues/43922 (filed by our very own @jcoyne 🙂).

i tried adding this (otherwise extraneous) `request.session` (inspired by [jcoyne's debugging here](https://github.com/rails/rails/issues/43922#issuecomment-999751892)), and that made the unexpected error go away, even when i let my test run the actual `Honeybadger.notify` call without mocking it.

totally open to suggestions for other workarounds for this issue.  the other obvious option from that thread seemed to be monkeypatching `Rack::Session::Abstract::PersistedSecure::SecureSessionHash`, which felt worse to me than what i did here.

app/controllers/dor_controller.rb
```diff
@@ -2,17 +2,21 @@
 
 # Main controller of application
 class DorController < ApplicationController
+  include Dry::Monads[:result]
+
+  class CocinaModelBuildError < StandardError; end
```
jmartin-sul, line 7: i wasn't sure if wrapping the build error was overkill, but it made it easier to keep the building in a private helper method and the error handling in the `DorController#reindex` method, while still being sure that only a build error was being rescued (as opposed to rescuing all of `StandardError` in `#reindex`).

spec/requests/dor_requests_spec.rb
```diff
@@ -54,6 +54,46 @@
...
+      it 'requires both the cocina json and the created_at/updated_at metadata' do
+        post "/dor/reindex/#{druid}", params: { cocina_model_json: cocina_model_json, updated_at: updated_at }
+        expect(response.code).to eq '200'
+        expect(mock_solr_conn).to have_received(:add).with(mock_solr_doc, add_attributes: { commitWithin: 1000 })
+        expect(object_service).to have_received(:find_with_metadata)
+      end
+
+      it 'provides the caller with a useful error if invalid cocina is provided' do
+        # Cocina::Models.build will be called with our bogus JSON, which will throw an error
+        allow(Honeybadger).to receive(:notify).and_call_original
```
jmartin-sul, line 86: normally i wouldn't tack on `.and_call_original` here, but i wanted to make sure that the hack for addressing https://github.com/rails/rails/issues/43922 got exercised (see other review comment).

i just confirmed that if i leave this line as is, but take out the `request.session` in `DorController#reindex`, the test fails (if i just `allow` with no call to the original method, the error is papered over and the test passes with or without the hack).

jcoyne (approved):

PR #770, opened 2022-03-03
OpenAPI schema: add new cocina model related params for /dor/reindex

jmartin-sul:

as @mjgiarlo pointed out, updates to the openapi.yml got missed in #769.

this seemed like the bare minimum, in that it adds the new optional params.

things that might be nice improvements, but that i wasn't sure how to implement (happy to read some docs and/or pair):
* schema for each of the new parameters that better constrains what values can be given for them.  i was unsure whether that'd require copypasta from another project's openapi.yml, or if there was something shared/centralized i could reference?
* ~~requirement for either all three new params or none of them.  i feel like i've seen a PR against one of our repos that enforced something like this via openapi.yml?~~ ended up taking care of this in the course of moving the new params to the request body.

also mjgiarlo and i were both slightly surprised that the new tests on #769 passed without an openapi.yml change 🤷 

unit tests

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -57,28 +57,31 @@ paths:
           required: false
           schema:
             type: integer
-        - name: cocina_object
-          in: query
-          description: the JSON serialization of the Cocina object
-          required: false
-          content:
-            application/json:
-              schema:
-                type: object
-        - name: created_at
-          in: query
-          description: the creation date of the Cocina object
-          required: false
-          schema:
-            type: string
-            format: date-time
-        - name: updated_at
-          in: query
-          description: the most recent modification date of the Cocina object
-          required: false
-          schema:
-            type: string
-            format: date-time
+      requestBody:
+        required: false
+        content:
+          '*/*':
+            schema:
+              type: object
+          application/json:
+            schema:
+              type: object
+              required:
+                - cocina_object
+                - created_at
+                - updated_at
+              properties:
+                cocina_object:
+                  description: JSON serialization of the Cocina object.  to be used, created_at and updated_at must also be provided.
+                  type: string
+                created_at:
+                  description: the creation date of the Cocina object
+                  type: string
+                  format: date-time
+                updated_at:
+                  description: the most recent modification date of the Cocina object
+                  type: string
+                  format: date-time
       responses:
         '200':
           description: Object successfully reindexed
--- a/openapi.yml
+++ b/openapi.yml
@@ -57,6 +57,28 @@ paths:
           required: false
           schema:
             type: integer
+        - name: cocina_model_json
+          in: query
+          description: the JSON serialization of the Cocina object
+          required: false
+          content:
+            application/json:
+              schema:
+                type: object
+        - name: created_at
+          in: query
+          description: the creation date of the Cocina object
+          required: false
+          schema:
+            type: string
+            format: date-time
+        - name: updated_at
+          in: query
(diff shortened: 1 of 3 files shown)


jmartin-sul:
undrafting in case it's worth merging this as-is and improving in a follow-on PR...

openapi.yml
```diff
@@ -57,6 +57,26 @@ paths:
           required: false
           schema:
             type: integer
+        - name: cocina_model_json
+          in: query
```
justinlittman, line 61: Are the params provided in query string or request body?

openapi.yml
```diff
@@ -57,6 +57,26 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                type: object
+        - name: created_at
+          in: query
+          description: the creation date of the Cocina object
+          required: false
+          schema:
+            type: string
```
justinlittman, line 73: `format: date-time`?

openapi.yml
```diff
@@ -57,6 +57,26 @@ paths:
...
+        - name: updated_at
+          in: query
+          description: the most recent modification date of the Cocina object
+          required: false
+          schema:
+            type: string
```
justinlittman, line 79: `format: date-time`?

openapi.yml
```diff
@@ -57,6 +57,26 @@ paths:
           required: false
           schema:
             type: integer
+        - name: cocina_model_json
```
justinlittman, line 60: Can we just call this `cocina` or `object` or `cocina_object`?

openapi.yml
jmartin-sul: thanks, was looking for something like that...

openapi.yml
jmartin-sul: i went with `cocina_object` for the param name, and shortened the corresponding var name in the app and test code (but left `_json` in that name since the app code might refer to either the built model instance or the JSON that's parsed to make it).

openapi.yml
jmartin-sul: in this case, they are provided in the query string.  i noticed when working on this PR that the cocina JSON is given in the request body for other API endpoints.  happy to make that change, i don't feel strongly about putting it in the query string, i just hadn't realized there was a slightly different existing pattern when working on #769 (and that didn't get called out in review of that one).

jmartin-sul: ok, changing after chatting w/ jcoyne a bit, as he pointed out that query size limits apply even though this is a POST...

jmartin-sul: ok, just pushed an update with all the `cocina_model` related params moved to the request body

jmartin-sul:
@justinlittman @jcoyne @mjgiarlo i think this is ready for another look, if one of you wants to give it another review.  thanks!

openapi.yml
```diff
@@ -57,6 +57,31 @@ paths:
           required: false
           schema:
             type: integer
+      requestBody:
+        required: false
+        content:
+          '*/*':
```
jmartin-sul, line 63: i found that i needed this section for backwards compatibility with the existing `/dor/reindex` usage

mjgiarlo: @jmartin-sul and I chatted about this and we think it's worth seeing if flipping (Committee gem config) `parse_response_by_content_type` to `false` (in `config/application.rb`) can suffice to allow the old behavior without allowing all content types for an optional body.

justinlittman (approved):

PR #775, opened 2022-03-04
DorController: split ability for caller to provide cocina out into new method

jmartin-sul:

extract the ability for the caller to provide cocina from `#reindex` into `#reindex_from_cocina`, per discussion on slack about #770.

this lets us get rid of the weird specification of `*/*` as a possible content type for the request body in situations where no content type is specified because no request body is provided by the caller (@mjgiarlo suspects a bug in the committee gem, as that was needed for backward compatibility with the old `/dor/reindex` route's existing usage, otherwise committee would throw an error about the expectation that content type must be `application/json`).

see [here](https://github.com/sul-dlss/dor_indexing_app/pull/770#discussion_r819171190) and [here](https://stanfordlib.slack.com/archives/C09M7P91R/p1646353580348509?thread_ts=1646268578.925449&cid=C09M7P91R).

unit tests

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -7,12 +7,19 @@ class DorController < ApplicationController
   class CocinaModelBuildError < StandardError; end
 
   def reindex
-    @indexer = Indexer.new(solr: solr, pid: params[:pid])
-
-    reindex_pid
+    cocina_with_metadata = indexer.fetch_model_with_metadata
+    reindex_pid(cocina_with_metadata)
     render status: :ok, plain: "Successfully updated index for #{params[:pid]}"
   rescue Dor::Services::Client::NotFoundResponse, Rubydora::RecordNotFound
     render status: :not_found, plain: 'Object does not exist in the repository'
+  end
+
+  def reindex_from_cocina
+    cocina_with_metadata = build_model_and_metadata(cocina_json: params[:cocina_object].presence,
+                                                    created_at: params[:created_at].presence,
+                                                    updated_at: params[:updated_at].presence)
+    reindex_pid(cocina_with_metadata)
+    render status: :ok, plain: "Successfully updated index for #{params[:pid]}"
   rescue CocinaModelBuildError => e
     request.session # TODO: calling this as a hack to address bad Rails/HB interaction, remove when https://github.com/rails/rails/issues/43922 is fixed
     Honeybadger.notify('Error building Cocina model', context: { druid: params[:pid], build_error: e.cause.message }, backtrace: e.cause.backtrace)
@@ -27,8 +34,6 @@ class DorController < ApplicationController
 
   private
 
-  attr_reader :indexer
-
   def solr
     RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solrizer_url)
   end
@@ -41,17 +46,11 @@ class DorController < ApplicationController
     raise CocinaModelBuildError
   end
 
-  # @returns [Success,Failure] the result of finding/parsing the model with metadata
-  def cocina_with_metadata
-    cocina_json = params[:cocina_object].presence
-    created_at = params[:created_at].presence
-    updated_at = params[:updated_at].presence
-    return indexer.fetch_model_with_metadata unless cocina_json && created_at && updated_at
-
-    build_model_and_metadata(cocina_json: cocina_json, created_at: created_at, updated_at: updated_at)
+  def indexer
+    @indexer ||= Indexer.new(solr: solr, pid: params[:pid])
   end
 
-  def reindex_pid
+  def reindex_pid(cocina_with_metadata)
     @solr_doc = indexer.reindex_pid(
       add_attributes: { commitWithin: params.fetch(:commitWithin, 1000).to_i },
       cocina_with_metadata: cocina_with_metadata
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -6,6 +6,7 @@
   namespace :dor do
     # TODO: Deprecate GET when caml POST can be implemented
     match 'reindex/:pid', action: :reindex, via: %i[get post put]
+    match 'reindex_from_cocina/:pid', action: :reindex_from_cocina, via: %i[get post put]
```
jcoyne, line 9: `get` and `post` should not be options here.  This is a PUT: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT

openapi.yml
```diff
@@ -57,12 +57,42 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+  /dor/reindex_from_cocina/{pid}:
+    post:
```
jcoyne, line 80: make this a PUT please.

spec/requests/dor_requests_spec.rb
```diff
@@ -23,39 +23,41 @@
...
+        allow(object_service).to receive(:find_with_metadata).and_raise(Dor::Services::Client::NotFoundResponse)
+        allow(connection).to receive(:find).and_raise(Rubydora::RecordNotFound)
+        post "/dor/reindex/#{druid}"
+        expect(response.body).to eq 'Object does not exist in the repository'
+        expect(response.code).to eq '404'
+      end
     end
 
-    context 'when cocina is provided by caller' do
+    describe 'POST #reindex_from_cocina' do
```
jcoyne, line 60: PUT

spec/requests/dor_requests_spec.rb
```diff
@@ -67,7 +69,7 @@
 
       it 'uses the provided cocina without hitting dor-services-app' do
         allow(Cocina::Models).to receive(:build).with(JSON.parse(cocina_json)).and_return(cocina) # pretend our bogus test JSON is valid
-        post "/dor/reindex/#{druid}",
+        post "/dor/reindex_from_cocina/#{druid}",
```
jcoyne, line 72: put

spec/requests/dor_requests_spec.rb
```diff
@@ -76,8 +78,19 @@
...
+             headers: { 'CONTENT_TYPE' => 'application/json' }
+        expect(response.body).to eq "Successfully updated index for #{druid}"
+        expect(response.code).to eq '200'
+        expect(mock_solr_conn).to have_received(:add).with(mock_solr_doc, add_attributes: { commitWithin: 10_000 })
+        expect(mock_solr_conn).not_to have_received(:commit)
+      end
+
       it 'requires both the cocina json and the created_at/updated_at metadata' do
-        post "/dor/reindex/#{druid}",
+        post "/dor/reindex_from_cocina/#{druid}",
```
jcoyne, line 93: put

spec/requests/dor_requests_spec.rb
```diff
@@ -76,8 +78,19 @@
         expect(object_service).not_to have_received(:find_with_metadata)
       end
 
+      it 'reindexes an object with specified commitWithin param and no hard commit' do
+        allow(Cocina::Models).to receive(:build).with(JSON.parse(cocina_json)).and_return(cocina) # pretend our bogus test JSON is valid
+        post "/dor/reindex_from_cocina/#{druid}",
```
jcoyne, line 83: put

spec/requests/dor_requests_spec.rb
```diff
@@ -87,7 +100,7 @@
       it 'provides the caller with a useful error if invalid cocina is provided' do
         # Cocina::Models.build will be called with our bogus JSON, which will throw an error
         allow(Honeybadger).to receive(:notify).and_call_original
-        post "/dor/reindex/#{druid}",
+        post "/dor/reindex_from_cocina/#{druid}",
```
jcoyne, line 103: put

jcoyne (changes requested):

app/controllers/dor_controller.rb
```diff
@@ -27,7 +34,7 @@ def delete_from_index
 
   private
 
-  attr_reader :indexer
+  attr_reader :cocina_with_metadata
```
jcoyne, line 37: How about making `reindex_pid` a function that takes `cocina_with_metadata` as a parameter rather than having an ivar.  This removes any state related bugs.

jcoyne (changes requested):

app/controllers/dor_controller.rb
jmartin-sul: sure, i was on the fence about that, will change

config/routes.rb
jmartin-sul: d'oh, yup, fixing...

jmartin-sul:
ok, changed the new route to accept `PUT` only, and turned `cocina_with_metadata` into a local var and param from an ivar in `DorController`.  ready for another look, i think.

jcoyne (approved):

PR #803, opened 2022-03-23
/dor/reindex_from_cocina can parse druid from json, remove redundant druid from URL path

jmartin-sul:

realized having the druid in the URL path was redundant and introduced opportunity for confusion if druid isn't parseable from cocina or if it differs.  see https://github.com/sul-dlss/dor-services-app/pull/3683#discussion_r822069566

connects with sul-dlss/dor-services-app#3651

- [x] unit test
- [x] integration tests

ran integration tests against QA with sul-dlss/dor-services-app#3683 and sul-dlss/dor_indexing_app#803 deployed, all passed.

### DIFF
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -18,12 +18,13 @@ class DorController < ApplicationController
     cocina_with_metadata = build_model_and_metadata(cocina_json: params[:cocina_object].presence,
                                                     created_at: params[:created_at].presence,
                                                     updated_at: params[:updated_at].presence)
-    reindex_object(cocina_with_metadata)
-    render status: :ok, plain: "Successfully updated index for #{params[:id]}"
+    druid = cocina_with_metadata.first.externalIdentifier
+    reindex_object(Success(cocina_with_metadata))
+    render status: :ok, plain: "Successfully updated index for #{druid}"
   rescue CocinaModelBuildError => e
     request.session # TODO: calling this as a hack to address bad Rails/HB interaction, remove when https://github.com/rails/rails/issues/43922 is fixed
-    Honeybadger.notify('Error building Cocina model', context: { druid: params[:id], build_error: e.cause.message }, backtrace: e.cause.backtrace)
-    render status: :unprocessable_entity, plain: "Error building Cocina model for #{params[:id]}"
+    Honeybadger.notify('Error building Cocina model', context: { cocina: params[:cocina_object], build_error: e.cause.message }, backtrace: e.cause.backtrace)
+    render status: :unprocessable_entity, plain: "Error building Cocina model from json: #{params[:cocina_object]}"
   end
 
   def delete_from_index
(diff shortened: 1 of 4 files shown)


app/controllers/dor_controller.rb
```diff
@@ -34,14 +36,24 @@ def delete_from_index
 
   private
 
+  def cocina_hash
+    @cocina_hash ||=
+      if params[:cocina_object].present?
+        # params[:cocina_object] is itself an ActionController::Parameters instance, #permit! is needed to allow the #to_h
+        params[:cocina_object].permit!.to_h
```
jcoyne, line 43:
```suggestion
        params[:cocina_object].to_unsafe_h
```

jcoyne (changes requested):

app/controllers/dor_controller.rb
```diff
@@ -34,14 +36,24 @@ def delete_from_index
...
+      else
+        {}
```
jcoyne, line 45: This is going to result in an error when you try to pass it to the builder, so why not just skip this conditional and do `params.require(:cocina_object)`?

jcoyne (changes requested):

app/controllers/dor_controller.rb
jmartin-sul: ah yeah, good point

app/controllers/dor_controller.rb
jmartin-sul: did this manually with your other suggestion and amended the most recent commit w/ both changes

jmartin-sul:
ready for another look:

```diff
dor_indexing_app % git diff origin/reindex_from_cocina-no-druid-in-url-path HEAD
diff --git a/app/controllers/dor_controller.rb b/app/controllers/dor_controller.rb
index 233af22..f09f5b3 100644
--- a/app/controllers/dor_controller.rb
+++ b/app/controllers/dor_controller.rb
@@ -15,6 +15,8 @@ class DorController < ApplicationController
   end
 
   def reindex_from_cocina
+    # params[:cocina_object] is itself an ActionController::Parameters instance, hence the #require and #to_unsafe_h
+    cocina_hash = params.require(:cocina_object).to_unsafe_h
     cocina_with_metadata = build_model_and_metadata(cocina_hash: cocina_hash,
                                                     created_at: params[:created_at].presence,
                                                     updated_at: params[:updated_at].presence)
@@ -36,16 +38,6 @@ class DorController < ApplicationController
 
   private
 
-  def cocina_hash
-    @cocina_hash ||=
-      if params[:cocina_object].present?
-        # params[:cocina_object] is itself an ActionController::Parameters instance, #permit! is needed to allow the #to_h
-        params[:cocina_object].permit!.to_h
-      else
-        {}
-      end
-  end
-
   def solr
     RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solrizer_url)
   end
```

app/controllers/dor_controller.rb
```diff
@@ -15,15 +15,19 @@ def reindex
   end
 
   def reindex_from_cocina
-    cocina_with_metadata = build_model_and_metadata(cocina_json: params[:cocina_object].presence,
+    # params[:cocina_object] is itself an ActionController::Parameters instance, hence the #require and #to_unsafe_h
```
jcoyne, line 18: I think this is fine, but I also think that it is well known (and well documented within Rails) that params in a Rails controller returns AC::Parameters

jcoyne (approved):

app/controllers/dor_controller.rb
jmartin-sul: ah yeah, i wasn't totally sure about putting that comment in, but i got tripped up for a bit by `params[:cocina_object]` returning a `Parameters` instance while the two date values just returned strings, so i thought it might be helpful when coming back to the code.

PR #883, opened 2022-08-02
Guard against subjects with no value

### LINKED ISSUE #882: NoMethodError: undefined method `sub' for nil:NilClass
This happens when indexing descriptive metadata that have a subject with no value. For example [druid:tf486tz1347](https://argo.stanford.edu/items/druid:tf486tz1347.json)

## Backtrace

    line 52 of [PROJECT_ROOT]/app/services/topic_builder.rb: flat_topic
    line 16 of [PROJECT_ROOT]/app/services/topic_builder.rb: block in build
    line 16 of [PROJECT_ROOT]/app/services/topic_builder.rb: each

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49898/faults/87092365)

edsu:

While we don't want Cocina to have subjects that have no value, we also don't want dor_indexing_app to blow up when it is sent them. This change will simply not try to remove punctuation from subjects that lack a value.

Do we know if Cocina validation should have trapped these earlier?

Fixes #882

Added a new unit test, and ran all of them.

### DIFF
--- a/app/services/topic_builder.rb
+++ b/app/services/topic_builder.rb
@@ -49,7 +49,7 @@ class TopicBuilder
       value.parallelValue.flat_map { |topic| flat_topic(topic) }
     elsif remove_trailing_punctuation?
       # comma, semicolon, and backslash are dropped
-      Array(value.value.sub(/[ ,;\\]+$/, ''))
+      Array(value.value&.sub(/[ ,;\\]+$/, ''))
     else
       Array(value.value)
     end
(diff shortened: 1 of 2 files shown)


app/services/topic_builder.rb
```diff
@@ -47,7 +47,7 @@ def parallel_subjects(subjects)
   def flat_topic(value)
     if value.parallelValue.present?
       value.parallelValue.flat_map { |topic| flat_topic(topic) }
-    elsif remove_trailing_punctuation?
+    elsif remove_trailing_punctuation? && value.value
       # comma, semicolon, and backslash are dropped
       Array(value.value.sub(/[ ,;\\]+$/, ''))
```
justinlittman, line 52: For your consideration. A slightly more idiomatic way to handle this would be to use the safe navigation operator: `Array(value.value&.sub(/[ ,;\\]+$/, ''))`

justinlittman (approved):
Approve pending consideration of comment.

PR #886, opened 2022-08-08
desc_metadata_indexer: avoid error when genre value is nil

ndushay:

To address https://app.honeybadger.io/projects/49898/faults/87288421

While we don't want Cocina to have genres that have no value, we also don't want dor_indexing_app to blow up when it is sent them. This change will simply not try to capitalize a genre that lacks a value.

@arcadiafalcone not sure how to re-index all the affected items ...

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -112,7 +112,7 @@ class DescriptiveMetadataIndexer
 
     return resource_type_formats if resource_type_formats == ['Book']
 
-    genre_formats = flat_forms_for('genre').map { |form| form.value.capitalize }.uniq
+    genre_formats = flat_forms_for('genre').map { |form| form.value&.capitalize }.uniq
 
     (resource_type_formats + genre_formats).presence
   end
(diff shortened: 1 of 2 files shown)


ndushay:
tagging @arcadiafalcone to make sure she's okay wth this.  Seems like a new error probably for one set of objects ...

jcoyne (changes requested):
Can you please add a test that demonstrates how this change fixes the problem?

arcadiafalcone:
@ndushay This error is possible when creating cocina without routing through MODS first. This also relates to the conversation about requiring certain properties in validation to catch orphan attributes without a corresponding value or value-equivalent property, which would be a more comprehensive way of addressing the issue.

Slack thread from when this was done for subject: https://stanfordlib.slack.com/archives/C09M7P91R/p1659439518438329

ndushay:
@jcoyne test added.

@arcadiafalcone I'll assume you'll make a ticket or whatever to address this problem with validation.

ndushay:
@arcadiafalcone also assume you or someone will reindex all the affected items?  (fewer than 30, I think, per honeybadger alert.)

arcadiafalcone:
Ticket for the general issue is here: https://github.com/sul-dlss/cocina-models/issues/510

arcadiafalcone:
I've added required property lists to https://github.com/sul-dlss/cocina-models/issues/510.

jcoyne (approved):

PR #903, opened 2022-09-19
Remove has_model_ssim

jcoyne:
It is no longer used.
Ref https://github.com/sul-dlss/argo/pull/3852

ci


app/indexers/data_indexer.rb
```diff
@@ -18,10 +18,10 @@ def to_solr
       solr_doc['modified_latest_dttsi'] = modified_latest
       solr_doc['created_at_dttsi'] = created_at
 
-      # These are required as long as dor-services-app uses ActiveFedora for querying:
-      solr_doc['has_model_ssim'] = legacy_model
```
mjgiarlo, line 22: Any reason to leave `#legacy_model` hanging around below? It looks like this was the only call to that method.

mjgiarlo (approved):
Approve pending consideration of comment about a now (apparently) unused method.

app/indexers/data_indexer.rb
jcoyne: Thanks for catching this. I've removed it.

mjgiarlo (approved):

PR #933, opened 2023-02-14
Document dynamic field semantics & link to schema

mjgiarlo:

Discussed in a meeting today. More clarity is better.

N/A


ndushay (approved):
Thank you, Mike.

Issue #940, opened 2023-02-23
multiple ILS identifiers - make them searchable


ndushay:
Here are the SEARCH fields in the Argo Solr index that reference ILS identifiers
(from https://github.com/sul-dlss/argo/blob/main/app/helpers/blacklight_config_helper.rb)

- `dor_id_tesim` => [source_id_value, barcode, catkey].compact
    - Included as a common SEARCH field in Argo (BlacklightConfigHelper)
- `identifier_tesim` => prefixed_identifiers*
    - Included as a common SEARCH field in Argo (BlacklightConfigHelper)  (here it is prefixed!)
    - Displayed in Argo as IDs field in search results(CatalogController) <-- ALERT, ALERT!

*prefixed_identifiers is a method that supplies, e.g. [naomi:test_record1, catkey:666].  Note that currently, prefixed_identifiers method also populates  `identifier_ssim` which is NOT a search field

Recent PR for these fields: #939

comment from @andrewjbtw in ticket #930:

>Having a multiple values will have a downstream effect on the Argo report view, which only returns the one index catkey. I'm not sure how best to return multiple values in the CSV but that's for another ticket.

**The work for this ticket is**:

First PR: #950
- [x] `dor_id_tesim` field should get any and all ILS identifiers in the cocina - all catkeys (not just the first one) and all folio instance hrids.
    - to test this fully:  deploy to -stage or -qa, find an Argo record with multiple ckeys (ask for help from Andrew or Arcadia? or find a virtual object from our integration tests????).  Re-index the record from Argo.  Ensure you get it in search results from each identifier.  Ensure the Search result and the Show page for the record looks okay.  Ask Andrew and Arcadia to check the record also -- many eyes, report view, and other goodness.

Second PR:
- [ ] `identifier_tesim`  field should get any and all ILS identifiers in the cocina - all catkeys (not just the first one) and all folio instance hrids.
    - currently uses prefixed_identifiers is a method that supplies, e.g. [naomi:test_record1, catkey:666];  however, we now want to have one consumer of prefixed_identifiers use ALL values of ILS identifiers, and the other consumer to stay as it is now.
[...]

PR #950, opened 2023-03-10
dor_id_tesim now includes previous ils ids

ndushay:

Part of #940 - this PR will allow Argo users to search on the value of identifiers (i.e. no "catkey" prefix) that are in cocina as "previous symphony" or "previous folio" and have them be included in the search results.   Note that `dor_id_tesim` is ONLY a search field in Argo and does not display in search results or in the record view -- you have to look at the cocina to see it.

Tagging @andrewjbtw as he should sign off on this approach.  As of writing this PR, it is deployed to qa and to stage.  You will likely need to manually reindex records to see them in search results.

unit test and did a test on qa where there were 2 search results for an id, and I reindexed a cocina object with that id as a previous symphony one, and voila - there were 3 search results.


app/indexers/identity_metadata_indexer.rb
```diff
@@ -41,10 +41,24 @@ def catkey
...
     @folio_instance_hrid ||= Array(cocina_object.identification.catalogLinks).find { |link| link.catalog == 'folio' }&.catalogRecordId
   end
 
+  def previous_folio_instance_hrids
+    @previous_folio_instance_hrids ||=
+      Array(cocina_object.identification.catalogLinks).filter_map { |link| link.catalogRecordId if link.catalog == 'previous folio' }
+  end
+
+  def previous_ils_ids
+    @previous_ils_ids ||= Array(previous_catkeys + previous_folio_instance_hrids)
```
mjgiarlo, line 59: Since both `#previous_catkeys` and `#previous_folio_instance_hrids` both are guaranteed to return an array, I think the `Array()` here should not be necessary?

justinlittman (approved):
Approve pending consideration of Mike's comment.

app/indexers/identity_metadata_indexer.rb
ndushay: I tried it without and it was unhappy.  :-(

ndushay: IIRC

mjgiarlo: @ndushay As in, specs were failing? Or did the badness manifest in some other way?

ndushay: I was confused -- I got the error in `catkey` and `folio_instance_id` methods

ndushay: (so now fixed)

mjgiarlo: 🍻

PR #960, opened 2023-05-03
Remove NewRelic dependency and deployment steps

mjgiarlo:

DIA does not use NewRelic and we do not want it.

CI

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -10,7 +10,7 @@ ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call
 set :deploy_to, '/opt/app/dor_indexer/dor_indexing_app'
 
 set :linked_dirs, %w[log config/settings tmp/pids tmp/cache tmp/sockets vendor/bundle]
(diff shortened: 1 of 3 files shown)


ndushay (approved):

config/deploy.rb
```diff
@@ -10,7 +10,7 @@
 set :deploy_to, '/opt/app/dor_indexer/dor_indexing_app'
 
 set :linked_dirs, %w[log config/settings tmp/pids tmp/cache tmp/sockets vendor/bundle]
-set :linked_files, %w[config/secrets.yml config/honeybadger.yml config/newrelic.yml]
+set :linked_files, %w[config/honeybadger.yml]
```
ndushay, line 13: is config/secrets.yml in shared_configs?  should it be removed if so?

ndushay (approved):

config/deploy.rb
mjgiarlo: It was removed as part of the secrets-to-vault mini-patch party.

PR #986, opened 2023-08-29
index dois and orcidids

peetucket:

Part of https://github.com/sul-dlss/argo/issues/4096

This will index DOIs (for items) and ORCIDIDs (for contributors) into solr.  

There will then be a separate Argo PR to add the appropriate facet (e.g. 'Identifiers', which then has multiple values, searching for any objects where the DOI/ORCID fields are *not* null).

~~HOLD pending test in QA or stage~~

New tests

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -8,6 +8,7 @@ class DescriptiveMetadataIndexer
     @cocina = cocina
   end
 
+  # rubocop:disable Metrics/MethodLength
   # @return [Hash] the partial solr document for descriptive metadata
   def to_solr
     {
@@ -16,6 +17,7 @@ class DescriptiveMetadataIndexer
       'sw_format_ssim' => sw_format,
       'sw_genre_ssim' => display_genre,
       'sw_author_tesim' => author,
+      'contributor_orcids_ssim' => orcids,
       'sw_display_title_tesim' => title,
       'sw_subject_temporal_ssim' => subject_temporal,
       'sw_subject_geographic_ssim' => subject_geographic,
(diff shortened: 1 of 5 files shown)


app/indexers/descriptive_metadata_indexer.rb
```diff
@@ -199,5 +207,20 @@ def flat_event(event)
...
+
+  # @param [Cocina::Models::Contributor] contributor to check
+  # @return [String, nil] orcid id including host if present
+  def orcidid(contributor)
+    identifier = contributor.identifier.find { |id| id.type == 'ORCID' }
+    return unless identifier
+
+    URI.join(identifier.source.uri, identifier.value).to_s
+  end
 end
```
peetucket, line 225: NOTE: these methods to search for ORCIDs were lifted from https://github.com/sul-dlss/orcid_client/blob/main/lib/sul_orcid_client/cocina_support.rb

While it would be nice to not copy and paste them here, that would require either (1) adding the sul-orcid-client gem has a dependency here or (2) moving these methods somewhere else that is available to both this app and the sul-orcid-client gem.  Currently opting for neither.

app/indexers/identity_metadata_indexer.rb
```diff
@@ -46,6 +47,10 @@ def previous_catkeys
       Array(cocina_object.identification.catalogLinks).filter_map { |link| link.catalogRecordId if link.catalog == 'previous symphony' }
   end
 
+  def doi
```
justinlittman, line 50: Why not just: `@doi ||= cocina_object&.identification.doi`?

justinlittman (changes requested):
This class is already too large and the ORCID handling adds enough code, that I would request that it be separated out into its own class. (I think we've been using the convention of calling those builders.)

(No issue with the code, though.)

peetucket:
> This class is already too large and the ORCID handling adds enough code, that I would request that it be separated out into its own class. (I think we've been using the convention of calling those builders.)
> 
> (No issue with the code, though.)

Abstracted ORCID code into a new class (using same pattern as `AuthorBuilder`) to reduce size of descriptive indexer class.

justinlittman (approved):

PR #989, opened 2023-09-06
deal with a different orcid format

peetucket:

This HB alert was triggered when trying to index records with a different ORCID format in the descriptive cocina than we were expecting.  This allows us to deal with it.  

see https://app.honeybadger.io/projects/49898/faults/99952126 and https://argo-prod-02.stanford.edu/view/druid:gf852zt8324 for the example record

Updated spec to include the alternate format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,10 @@ $ cap ENV rolling_indexer:restart
 Or if you're on a server that has the `rolling_indexer` capistrano role, use systemd commands:
 
 ```shell
-$ sudo systemctl status rolling_index
-$ sudo systemctl start rolling_index
-$ sudo systemctl stop rolling_index
-$ sudo systemctl restart rolling_index
+$ sudo systemctl status rolling-index
+$ sudo systemctl start rolling-index
+$ sudo systemctl stop rolling-index
+$ sudo systemctl restart rolling-index
 ```
 
 **NOTE 1**: The rolling indexer is automatically restarted during deployments.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,10 +22,10 @@ $ cap ENV rolling_indexer:restart
...
 ```shell
-$ sudo systemctl status rolling_index
-$ sudo systemctl start rolling_index
-$ sudo systemctl stop rolling_index
-$ sudo systemctl restart rolling_index
+$ sudo systemctl status rolling-index
+$ sudo systemctl start rolling-index
+$ sudo systemctl stop rolling-index
+$ sudo systemctl restart rolling-index
 ```
```
peetucket, line 29: Changed to match DevOpsDocs: https://github.com/sul-dlss/DevOpsDocs/blob/master/projects/dor_indexing_app/operations-concerns.md and what the cap tasks actually do

justinlittman (approved):

app/services/orcid_builder.rb
```diff
@@ -38,6 +38,9 @@ def orcidid(contributor)
     identifier = contributor.identifier.find { |id| id.type == 'ORCID' }
     return unless identifier
 
+    # some records have the full ORCID URI in the data, just return it if so, e.g. druid:gf852zt8324
+    return identifier.uri if identifier.uri
+
```
edsu, line 43: This is a good example of something that could be simplified in Cocina.

peetucket:
@arcadiafalcone just FYI, when deploying a change to indexing code to add ORCIDs for contributors in descriptive metadata, we hit a druid that had a slightly different format than the one we are using for H2 records.  You can see it here: https://argo-prod-02.stanford.edu/view/druid:gf852zt8324

This completed PR for dor_indexing_app allows our indexing code to deal with this difference and work in both cases, but not sure if it was expected or not and if there will be other records like this (or possibly even different from H2 and this one).

Issue #992, opened 2023-09-14
Index argo facets starting "SW" with the same code used by SearchWorks for those facets


ndushay:
Use stanford_mods code to index the Argo facets that are supposed to be akin to SearchWorks facets.  There is an exemplar PR #993.  I think a single PR for all the rest of the facets below is fine, but that's up to you.

The below is condensed from https://docs.google.com/spreadsheets/d/1s4gQcPEY_YPGrDOFXQQ1IbRYBc3AaKGRIPW1FFw46G0/, where I traced Argo and SearchWorks for catalog config information for Solr facet field names and then chased indexing code in searchworks_traject_indexer to get to the methods used to populate the SearchWorks facet fields.  See also the epic sul-dlss/argo#3168 for more info.

All SW facets:
-  UI facet name | Argo Solr field | SW Solr field | stanford_mods method
- [ ] SW Resource type | sw_format_ssim | format_main_ssim | Stanford::Mods::Searchworks.format_main (PR #1005)
- [x] SW Date | sw_pub_date_facet_ssi | pub_year_tisim | Stanford::Mods::OriginInfo.pub_year_int (See PR #1008)
- [x] SW Topic | topic_ssim | topic_facet | Stanford::Mods::SearchworksSubjects.topic_facet (See PR #1009)
- [ ] SW Region | sw_subject_geographic_ssim | geographic_facet | Stanford::Mods::SearchworksSubjects.geographic_facet (PR #1010)
- [x] SW Era | sw_subject_temporal_ssim | era_facet | Stanford::Mods::SearchworksSubjects.era_facet (See PR #1011)
- [x] SW Genre | sw_genre_ssim | genre_ssim | Stanford::Mods::Searchworks.sw_genre (See PR #993)
- [x] SW Language | sw_language_ssim | language : sw_language_facet | Stanford::Mods::Searchworks.sw_language_facet (PR #1012)

ndushay:
The below is condensed from https://docs.google.com/spreadsheets/d/1s4gQcPEY_YPGrDOFXQQ1IbRYBc3AaKGRIPW1FFw46G0/, where I traced Argo and SearchWorks for catalog config information for Solr facet field names and then chased indexing code in searchworks_traject_indexer to get to the methods used to populate the SearchWorks facet fields.

All SW facets:
-  UI facet name : Argo Solr field : SW Solr field : stanford_mods method
[...]

PR #1010, opened 2023-09-25
Update sw_subject_geographic_ssim to use Stanford::Mods::SearchworksS…

aaron-collier:
…ubjects.geographic_facet

Part of #992 

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -24,7 +24,7 @@ class DescriptiveMetadataIndexer
       'contributor_orcids_ssim' => orcids,
       'sw_display_title_tesim' => title,
       'sw_subject_temporal_ssim' => subject_temporal,
-      'sw_subject_geographic_ssim' => subject_geographic,
+      'sw_subject_geographic_ssim' => stanford_mods_record.geographic_facet.uniq,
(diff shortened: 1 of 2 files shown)


app/indexers/descriptive_metadata_indexer.rb
```diff
@@ -24,7 +24,7 @@ def to_solr
       'contributor_orcids_ssim' => orcids,
       'sw_display_title_tesim' => title,
       'sw_subject_temporal_ssim' => subject_temporal,
-      'sw_subject_geographic_ssim' => subject_geographic,
+      'sw_subject_geographic_ssim' => stanford_mods_record.geographic_facet.uniq,
```
mjgiarlo, line 27: It looks like all the test changes below are negative tests now. What *does* the change now map to?

ndushay:
Specs make it look like "it's opposite day!" -- this seems to show significant logic differences between straight cocina indexing and stanford-mods indexing.

ndushay (changes requested):
do not merge until the world is a better place.

ndushay:
ACTION ITEM per discussion with @andrewjbtw @ndushay and @aaron-collier on 10/3/23:  Aaron will write "report" tickets for sussing out the differences between stanford mods indexing and dor-indexing-app for this part of descriptive metadata.

PR #1013, opened 2023-09-26
[HOLD | DRAFT] Update sw_author_tesim facet to use stanford-mods.sw_main_author

### LINKED ISSUE #994: Investigate adding the SW author facet to Argo, indexed with same code as SearchWorks Author facet
There is a desire to (re-)add a SW Author facet to Argo.  It was removed because folks were confused by the contents.   

Currently the results of https://github.com/sul-dlss/dor_indexing_app/blob/main/app/services/author_builder.rb are very different from https://github.com/sul-dlss/stanford-mods/blob/master/lib/stanford-mods/concerns/name.rb#L11. Initially investigate the data differences (add a report to DSA to extract the data used for comparison)

Check with @andrewjbtw if this should be solely for person Authors, as SearchWorks as a second facet for Author (organization) ... which could get similar treatment, or could be ignored, or could be combined into a single facet in Argo.  (My assumption is that SDR won't have a lot of organization authors we care about but what do I know?)

Use the same approach taken in #992:  (partial exemplar PR #993)

- determine which Solr field is used for the author facet in SearchWorks (probably via catalog_config or some such blacklight thing)
- determine from searchworks_traject_indexer how that field is indexed (see https://docs.google.com/spreadsheets/d/1s4gQcPEY_YPGrDOFXQQ1IbRYBc3AaKGRIPW1FFw46G0 ).
- if the indexing logic is encapsulated in the stanford_mods gem, as for the other SW facet fields in Argo, then proceed as PR #993 indicates to add indexing code here.
- after indexing code is merged and deployed, prepare PR that configures Argo to include the new facet.  Talk to Andrew about timing - should it appear before it is fully populated, etc.

If the course ahead is less straightforward, discuss with Naomi / Andrew about whether it's worth it.

aaron-collier:

Closes #994 

The different tests highlighted below should be reviewed by @andrewjbtw to discuss the usefulness of this change.

### DIFF
--- a/spec/indexers/descriptive_metadata_indexer_contributor_spec.rb
+++ b/spec/indexers/descriptive_metadata_indexer_contributor_spec.rb
@@ -162,7 +162,7 @@ RSpec.describe DescriptiveMetadataIndexer do
       end
 
       it 'selects primary name of contributor' do
-        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L.')
+        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L. Sayers, Dorothy L. (Dorothy Leigh), 1893-1957')
       end
     end
 
@@ -190,7 +190,7 @@ RSpec.describe DescriptiveMetadataIndexer do
       end
 
       it 'selects first name of contributor' do
-        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L.')
+        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L. Sayers, Dorothy L. (Dorothy Leigh), 1893-1957')
       end
     end
 
@@ -223,7 +223,7 @@ RSpec.describe DescriptiveMetadataIndexer do
       end
 
       it 'selects primary name from parallelValue' do
-        expect(doc).to include('sw_author_tesim' => 'Bulgakov, Mikhail Afanasʹevich')
+        expect(doc).to include('sw_author_tesim' => 'Булгаков, Михаил Афанасьевич')
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/indexers/descriptive_metadata_indexer_contributor_spec.rb
```diff
@@ -223,7 +223,7 @@
       end
 
       it 'selects primary name from parallelValue' do
-        expect(doc).to include('sw_author_tesim' => 'Bulgakov, Mikhail Afanasʹevich')
+        expect(doc).to include('sw_author_tesim' => 'Булгаков, Михаил Афанасьевич')
```
ndushay, line 226: it concerns us that this isn't the primary.

spec/indexers/descriptive_metadata_indexer_contributor_spec.rb
```diff
@@ -162,7 +162,7 @@
       end
 
       it 'selects primary name of contributor' do
-        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L.')
+        expect(doc).to include('sw_author_tesim' => 'Sayers, Dorothy L. Sayers, Dorothy L. (Dorothy Leigh), 1893-1957')
```
ndushay, line 165: these doubles are concerning.

ndushay (changes requested):
ACTION ITEM per discussion with @andrewjbtw @ndushay and @aaron-collier on 10/3/23:  Aaron will write "report" tickets for sussing out the differences between stanford mods indexing and dor-indexing-app for this part of descriptive metadata.

ndushay:
This isn't going to happen.

Issue #1006, opened 2023-09-25
index exploded non-project tags to NEW  to Solr field


ndushay:
This ticket is a follow on to #1004.

Late breaking news!  As of meeting on 10/4/22 (see https://docs.google.com/document/d/1n3jMSN-ywO8pyLEqsD0tNCsLR2y_p9tLAGVOvSSkbyQ/edit#heading=h.a0sj9f81gzds), the work for this ticket is to 

Write code to index (exploded) non-project tags into a NEW Solr field for all that is in the current exploded tags field (exploded_tags_ssim) except project tags.

--- old stuff below this line ---

Once the Solr field `exploded_project_tag_ssim` is fully (or nearly fully) populated (and the indexing of this new field went into production on Monday afternoon, 10/2, and it will take 2 weeks to fully index ...

... then stop writing project tags to `tag_ssim` or `exploded_tag_ssim` and only write them to `project_tag_ssim` and `exploded_project_tag_ssim`

Further note:  this work is blocked by the deploy of the new project tag facet coded in https://github.com/sul-dlss/argo/pull/4206 (https://github.com/sul-dlss/argo/issues/4187), which is blocked at the time of writing waiting for `exploded_project_tag_ssim` to be populated.   

So the PR for this work should NOT be (merged) deployed to prod before the new Project Tag facet displays in production Argo.

PR #1014, opened 2023-09-27
Exploded project tag indexing etc

### LINKED ISSUE #1004: index "project tag" into a separate field
Some argo power users search for matches on project tags, which all start "Project : " and are made that way in the Register Items form (see Project Name text field)

image

This ticket is to make a PR that:
- [x] writes project tags to a new Solr field, `project_tag_ssim`, probably in [app/indexers/administrative_tag_indexer.rb#L18](https://github.com/sul-dlss/dor_indexing_app/blob/main/app/indexers/administrative_tag_indexer.rb#L18).  Do NOT include "Project : " prefix.  (Note trailing space in prefix!)
    - [ ] probably use a private method `project_tag?` which could also be useful for ticket #1006
- [x] also writes project tag values to a new `exploded_project_tag_ssim`, analogous to `exploded_tag_ssim`
- [ ] once the update has been deployed, confirm that Solr documents reindexed contain the new field.

jmartin-sul:

fixes #1004

deviations from the suggestions in the ticket
* ticket suggests adding a private helper method `project_tag?`, but the tag splitting that's already there makes this feel a little unnecessary.  even as groundwork for #1006, which #1004 blocks, this feels a little unnecessary given what's already there
* ticket suggests writing "project tags to a new Solr field, `project_tag_ssim`"... but afaict, we are already indexing project tags that way?  i see it in the DIA solr doc building code and unit tests, and in the Solr admin schema view, and in use in the argo code.  so i just left that as-is.

per the ticket, i only indexed `exploded_project_tag`, but it'd be easy to do a similar sort of indexing for the other special tag, `Registered By`/`registered_by`.

- [x] unit tests/CI
- [x] deploy to stage or QA
- [x] examined solr doc for a QA object with project tags, and one without, confirmed that project and exploded project tag fields looked fine on former, and that they weren't present on latter

### DIFF
--- a/app/indexers/administrative_tag_indexer.rb
+++ b/app/indexers/administrative_tag_indexer.rb
@@ -27,7 +27,12 @@ class AdministrativeTagIndexer
       next if SPECIAL_TAG_TYPES_TO_INDEX.exclude?(tag_prefix) || rest.nil?
 
       prefix = tag_prefix.downcase.strip.gsub(/\s/, '_')
+
       (solr_doc["#{prefix}_tag_ssim"] ||= []) << rest.strip
+      if prefix == 'project'
+        solr_doc['exploded_project_tag_ssim'] ||= []
+        solr_doc['exploded_project_tag_ssim'] += exploded_tags_from(rest.strip)
+      end
     end
     solr_doc
   end
--- a/app/indexers/administrative_tag_indexer.rb
+++ b/app/indexers/administrative_tag_indexer.rb
@@ -5,7 +5,7 @@
 #       https://github.com/sul-dlss/dor-services/blob/v9.0.0/lib/dor/datastreams/identity_metadata_ds.rb#L196-L218
 class AdministrativeTagIndexer
   TAG_PART_DELIMITER = ' : '
-  TAGS_TO_INDEX = ['Project', 'Registered By'].freeze
+  SPECIAL_TAG_TYPES_TO_INDEX = ['Project', 'Registered By'].freeze
 
   attr_reader :id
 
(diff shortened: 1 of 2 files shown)


app/indexers/administrative_tag_indexer.rb
```diff
@@ -5,7 +5,7 @@
 #       https://github.com/sul-dlss/dor-services/blob/v9.0.0/lib/dor/datastreams/identity_metadata_ds.rb#L196-L218
 class AdministrativeTagIndexer
   TAG_PART_DELIMITER = ' : '
-  TAGS_TO_INDEX = ['Project', 'Registered By'].freeze
+  SPECIAL_TAG_TYPES_TO_INDEX = ['Project', 'Registered By'].freeze
```
jmartin-sul, line 8: opportunistic touchup in the first commit, because the old name struck me as misleadingly generic

app/indexers/administrative_tag_indexer.rb
```diff
@@ -24,10 +24,15 @@ def to_solr
       solr_doc['exploded_tag_ssim'] += exploded_tags_from(tag)
 
       tag_prefix, rest = tag.split(TAG_PART_DELIMITER, 2)
-      next if TAGS_TO_INDEX.exclude?(tag_prefix) || rest.nil?
+      next if SPECIAL_TAG_TYPES_TO_INDEX.exclude?(tag_prefix) || rest.nil?
 
       prefix = tag_prefix.downcase.strip.gsub(/\s/, '_')
+
       (solr_doc["#{prefix}_tag_ssim"] ||= []) << rest.strip
```
jmartin-sul, line 31: this is the line that was already doing the first part of what the ticket asked for

app/indexers/administrative_tag_indexer.rb
```diff
@@ -24,10 +24,15 @@ def to_solr
...
+      if prefix == 'project'
+        solr_doc['exploded_project_tag_ssim'] ||= []
+        solr_doc['exploded_project_tag_ssim'] += exploded_tags_from(rest.strip)
```
jmartin-sul, line 34: these lines do the other part of what the ticket asked for (i got a syntax error when i tried to do them as one line, i.e. `(solr_doc['exploded_project_tag_ssim'] ||= []) += exploded_tags_from(rest.strip)`)

jmartin-sul: also, i'd originally added indexing for exploded versions of both the project tag and the registered by tag.  but from a quick glance at the argo prod data, it didn't seem like registered by tags ever had multiple levels.  so an exploded version didn't seem particularly useful, and the ticket didn't ask for it, so i kept this change confined to project tags.

mjgiarlo (approved):

PR #1016, opened 2023-09-28
Remove Symphony from the metadata_source facet

### LINKED ISSUE #1015: Remove "Symphony" from the metadata source fact
Stop showing "Symphony" as a metadata source. It's all FOLIO's world now.

aaron-collier:

Closes #1015 

### DIFF
--- a/app/indexers/identifiable_indexer.rb
+++ b/app/indexers/identifiable_indexer.rb
@@ -3,6 +3,8 @@
 class IdentifiableIndexer
   attr_reader :cocina
 
+  CURRENT_CATALOG_TYPE = 'folio'
+
   def initialize(cocina:, **)
     @cocina = cocina
   end
@@ -48,7 +50,7 @@ class IdentifiableIndexer
             .map(&:catalog)
             .uniq
             .sort
-            .reject { |catalog_type| catalog_type.include?('previous') }
+            .select { |catalog_type| catalog_type == CURRENT_CATALOG_TYPE }
   end
 
   # @param [Hash] solr_doc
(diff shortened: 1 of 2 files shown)


app/indexers/identifiable_indexer.rb
```diff
@@ -48,7 +50,7 @@ def distinct_current_catalog_types
             .map(&:catalog)
             .uniq
             .sort
-            .reject { |catalog_type| catalog_type.include?('previous') }
+            .reject { |catalog_type| DEPRECATED_CATALOGS.any? { |catalog| catalog_type.include?(catalog) } }
```
mjgiarlo, line 53: There are four possible values we can have in this field:

1. `folio`
2. `previous folio`
3. `symphony`
4. `previous symphony`

And the code as currently written rejects 2-4 and allows 1. Which is good! Given "previous" isn't a catalog that's been deprecated, I wonder if we might consider an even more direct approach here:

```suggestion
            .select { |catalog_type| catalog_type == CURRENT_CATALOG_TYPE }
```

Which selects 1 without caring about 2-4, which means less baggage to hang onto.

app/indexers/identifiable_indexer.rb
```diff
@@ -3,6 +3,8 @@
 class IdentifiableIndexer
   attr_reader :cocina
 
+  DEPRECATED_CATALOGS = %w[symphony previous].freeze
```
mjgiarlo, line 6: Along with my suggestion below:

```suggestion
  CURRENT_CATALOG_TYPE = 'folio'
```

mjgiarlo (approved):

PR #1035, opened 2023-10-12
Index description text.

### LINKED ISSUE #1029: populate all_text_timv field with all useful human readable text in cocina descriptive
We need to add a new field to our Solr documents, all_text_timv,  that is a "catch all" field for any cocina descriptive text that users might search on.

This should include the values (not the keys) from our cocina descriptive JSON:
- displayLabel
- value

and of course, recurse through structuredValues, parallelValues and groupedValues.

do NOT recurse through or include
- access
- adminMetadata 
- encoding 
- marcEncodedData
- identifier
- purl
- source
- standard
- valueAt
- valueLanguage 
- valueScript

Do this for these fields in the Cocina:
- [ ] title
- [ ] contributor
- [ ] event
- [ ] form
- [ ] geographic
- [ ] language 
- [ ] note
- [ ] relatedResource
- [ ] subject

See https://github.com/sul-dlss/dor-services-app/issues/4522 for more info ... but we don't want to include "uri" or "valueAt" or "code" (which is also a URI)

justinlittman:
closes #1029

See ticket

Unit

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -31,7 +31,8 @@ class DescriptiveMetadataIndexer
       'originInfo_place_placeTerm_tesim' => event_place,
       'topic_ssim' => stanford_mods_record.topic_facet.uniq,
       'topic_tesim' => stemmable_topics,
-      'metadata_format_ssim' => 'mods' # NOTE: seriously? for cocina????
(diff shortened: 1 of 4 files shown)


ndushay (approved):
Nice job, Justin!  I knew you would come up with something elegant!

I am merging this because it will be a big improvement over what we have now.  I am also tagging @arcadiafalcone to see if she wants to refine the indexing strategy I pulled out of the air for #1029.

For example:  do we want the contents of all notes?   What about form values?

PR #1037, opened 2023-10-12
descriptive text catch all searching field changes

ndushay:

- Arcadia suggested NOT indexing displayLabel values
- it should be a string, not separate values, so phrase searching works for values broken into bits (titles, names, topics ...)
- it should be indexed into 3 fields with different text analysis algorithms (stemming, tokenization ...)

specs

### DIFF
--- a/app/services/all_search_text_builder.rb
+++ b/app/services/all_search_text_builder.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-# Extracts useful text from Cocina Description
-class AllTextBuilder
+# Extracts useful text for searching from Cocina Description
+class AllSearchTextBuilder
   def self.build(cocina_description)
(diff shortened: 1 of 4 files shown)


app/services/all_search_text_builder.rb
```diff
@@ -13,15 +13,15 @@ def initialize(cocina_description)
   def build
     @text = []
     recurse(cocina_description)
-    text.uniq
+    text.join(' ')
```
ndushay, line 16: we want the values concatenated into a single string so we can match phrases against cocina values broken up into parts, like names or titles or topics may be.

ndushay:
This does not need to be held for @justinlittman but I wanted to tag him in case he's interested.

peetucket (approved):

PR #1040, opened 2023-10-13
obj_label_tesim - comment added as to where it is used

ndushay:

Because this just bit us.

comment change only


justinlittman (changes requested):
It is used in other places as well.

ndushay:
nevermind - killing this PR

PR #1042, opened 2023-10-17
Parallelize the slowest part of the rolling indexer

mjgiarlo:
# Why was this change made?

The rolling indexer is slower than we think it should be. We currently estimate that a complete reindex of SDR takes roughly ~11 days to finish. Naomi and I spent some time benchmarking the indexer, and noticed two ways we could speed this up:

1. Reduce the number of times we add docs to Solr from 500 to 1, as `RSolr::Client#add` can take a batch of docs in an array (shaves ~10s off the run);
2. Parallelize the `Dor::Services::Client#find` and the `Indexer.load_and_index` calls, which is where the indexer was spending most of its time

As a result, a typical run in prod takes ~20 seconds instead of ~145 seconds.  This is a seven fold increase (based on our sample size of 1 run).   See below for benchmarking results.

We also tested varying the Solr `commitWithin` value and found that higher values did not yield a faster runtime, which pointed to the source of sluggishness being elsewhere (namely the DSC and Indexer calls mentioned above):

```
                             user     system      total        real
index 500 (1s commit)   58.971828   2.422009  61.393837 (144.080748)
index 500 (10s commit)  59.640639   2.644295  62.284934 (143.016791)
index 500 (100s commit) 59.226527   2.276780  61.503307 (144.232593)
```

## Before

```
                             user     system      total        real
index 500 (1s commit):  61.467601   2.203758  63.671359 (142.879938)
```

## After parallelization

```
                                   user     system      total        real
index 500 as array (parallel)  1.045042   0.097685  60.504973 ( 20.565835)
```

## After adding `Indexer.load_and_build`

```
                                     user     system      total        real
only add once (for real)         1.084629   0.090197  60.799066 ( 18.072022)
```

## How Slow Has It Been?

We currently have around 7M objects in the index, and we're indexing 500 at a time,  so this will take 7M/500 = 14,000 iterations.  If each iteration takes 1 minute to run, then it's 14,000 minutes = 233 hours = 10 days PLUS 5 seconds wait time for each iteration, which adds another 19.4 hours so ... basically 11 days.  Except in observation, it seems it takes longer than 2 weeks.  See https://github.com/sul-dlss/argo/issues/4187#issuecomment-1765451949 and https://github.com/sul-dlss/argo/issues/4194#issuecomment-1765450594.  In particular, the tags indexing should have been completed by now.

# How was this change tested?

- [x] CI
- [x] Ran the benchmarking in prod (without changing the current rolling indexer process)
- [x] Deployed to stage and watched honeybadger

### DIFF
--- a/bin/rolling_index
+++ b/bin/rolling_index
@@ -2,6 +2,7 @@
 
 require_relative '../config/environment'
 require 'daemons'
+require 'parallel'
 
 QUERY = { q: '*:*', sort: 'timestamp asc', fl: 'id', rows: '500' }
 
@@ -9,20 +10,26 @@ Daemons.run_proc('rolling index') do
   loop do
     solr_conn = RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solrizer_url)
     response = solr_conn.get 'select', params: QUERY
-    response['response']['docs'].each do |doc|
+    solr_docs = Parallel.map(response['response']['docs'], in_processes: 8) do |doc|
       identifier = doc['id'].scrub('')
       # Occasionally, we've seen invalid bytes in the identifier, so try to catch those:
       Honeybadger.notify("Identifier isn't valid utf-8", { identifier: identifier, bytes: identifier.unpack('C*') }) unless doc['id'].valid_encoding?
       begin
+        # This returns a Solr doc instance
         Indexer.load_and_index(solr: solr_conn, identifier: identifier)
       rescue Dor::Services::Client::NotFoundResponse
         Honeybadger.notify('Rolling indexer cannot reindex since not found.', { druid: identifier })
         # Clean up cruft in QA and stage
         Indexer.delete(solr: solr_conn, identifier: identifier) if ['stage', 'qa'].include?(ENV['HONEYBADGER_ENV'])
+        # Return `nil`, which is compacted, so the Solr add isn't grumpy
+        nil
       rescue Dor::Services::Client::UnexpectedResponse
         Honeybadger.notify('Unexpected response from Dor Services App.', { druid: identifier })
+        # Return `nil`, which is compacted, so the Solr add isn't grumpy
+        nil
       end
-    end
+    end.compact
+    solr_conn.add(solr_docs, add_attributes: { commitWithin: 1000 })
     sleep(5) # This was just a wild guess.  We can turn this up/down as necessary.
   end
 end
--- a/bin/rolling_index
+++ b/bin/rolling_index
@@ -16,7 +16,7 @@ Daemons.run_proc('rolling index') do
       Honeybadger.notify("Identifier isn't valid utf-8", { identifier: identifier, bytes: identifier.unpack('C*') }) unless doc['id'].valid_encoding?
       begin
         # This returns a Solr doc instance
-        Indexer.load_and_index(solr: solr_conn, identifier: identifier)
+        Indexer.load_and_build(identifier: identifier)
       rescue Dor::Services::Client::NotFoundResponse
         Honeybadger.notify('Rolling indexer cannot reindex since not found.', { druid: identifier })
         # Clean up cruft in QA and stage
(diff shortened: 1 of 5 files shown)


ndushay:
notifying @justinlittman and @andrewjbtw that the only negative impact we have thought of is that by parallelizing the rolling reindexer, there will be fewer cores on -a for on demand indexing (and new object indexing) ... but -b is unaffected.  So keep an eye out for unacceptable indexing times for stuff.

andrewjbtw:
Presumably now that we've left Fedora, the possible negative impact of taking the core part of the system down with too much activity is less of a risk.

bin/rolling_index
```diff
@@ -2,27 +2,34 @@
...
     solr_conn = RSolr.connect(timeout: 120, open_timeout: 120, url: Settings.solrizer_url)
     response = solr_conn.get 'select', params: QUERY
-    response['response']['docs'].each do |doc|
+    solr_docs = Parallel.map(response['response']['docs'], in_processes: 8) do |doc|
       identifier = doc['id'].scrub('')
       # Occasionally, we've seen invalid bytes in the identifier, so try to catch those:
       Honeybadger.notify("Identifier isn't valid utf-8", { identifier: identifier, bytes: identifier.unpack('C*') }) unless doc['id'].valid_encoding?
       begin
+        # This returns a Solr doc instance
         Indexer.load_and_index(solr: solr_conn, identifier: identifier)
```
jmartin-sul, line 19: it looks from the `Indexer.load_and_index` ultimately does a `solr_conn.add` underneath, so i'm wondering if that means this the `solr_conn.add` that was added in this PR is redundant?  https://github.com/sul-dlss/dor_indexing_app/blob/ad5ee6221adc0f81c2a906aded54b6f6bc73fd1a/app/services/indexer.rb#L35-L53

i couldn't quite tell (based on title and description) whether the goal of this PR was only to gain speed via the parallelization change, or to also gain speed via reduction of the number of solr doc adds.  if the latter, and if i'm understanding the code correctly, it seems like we'd maybe want to refactor `Indexer` to have a method that only returns the solr doc instead of also adding it, or to replace the `Indexer.load_and_index` call here with e.g. `DocumentBuilder.for(model: Dor::Services::Client.object(identifier).find).to_solr` (and if going that route, to maybe add similar `Honeybadger.context` setting to what would be lost if `Indexer` isn't used).

jmartin-sul (approved):
based on a quick trace through the code, i'm wondering if there's an additional easy win to be gained in terms of speed based on the speed improvement possibilities mentioned in the PR discussion, assuming my read of the code is correct.

even with that concern, my read is that this change doesn't hurt things in terms of correctness.  and given the huge speed improvement and the expectation that even parallelization alone could provide a big speed boost, i'm going to assume that the speed improvement is real and that the test result isn't a random fluke due to the sample size of 1.

so, approving, but leaving unmerged for consideration of all that, and maybe for other eyes to review.

(dropping in on this PR review without any slack or standup discussion context, since i've mostly been out today and yesterday for jury duty)

ndushay (approved):
Nice observation, @jmartin-sul !  w00t!

mjgiarlo:
@jmartin-sul 💬 
> based on a quick trace through the code, i'm wondering if there's an additional easy win to be gained in terms of speed based on the speed improvement possibilities mentioned in the PR description, assuming my read of the code is correct.

Yes, good call! Made that change. It's now an ~8x improvement. 😄 
 
> even with that concern, my read is that this change doesn't hurt things in terms of correctness. and given the huge speed improvement and the expectation that even parallelization alone could provide a big speed boost, i'm going to assume that the speed improvement is real and that the test result isn't a random fluke due to the sample size of 1.

Between yesterday and today, @ndushay and I tested this over ~15 times in stage and prod, so the sample size is slightly bigger and the results were pretty consistent, for what it's worth!
 
🍻

jmartin-sul (approved):

ndushay:
@mjgiarlo after your FR shift and the problems ... did you end up removing the parallelization (set it to 1) to avoid the errors for DSA/DIA stuff?  Or are we still at 2?

mjgiarlo:
@ndushay still 2: https://github.com/sul-dlss/dor_indexing_app/blob/main/config/settings.yml#L4

PR #1043, opened 2023-10-18
add contributor_text_nostem_im field for searching authors

ndushay:

Part of sul-dlss/argo#4231

For better searching recall and precision.  In argo prod right now, there is very poor recall for searching on author names.

CI. (we're just adding a new field to the index)

### DIFF
--- a/app/indexers/descriptive_metadata_indexer.rb
+++ b/app/indexers/descriptive_metadata_indexer.rb
@@ -20,7 +20,9 @@ class DescriptiveMetadataIndexer
       'mods_typeOfResource_ssim' => resource_type,
       'sw_format_ssim' => sw_format,
       'sw_genre_ssim' => stanford_mods_record.sw_genre,
-      'sw_author_tesim' => author,
+      'contributor_text_nostem_im' => author_all, # author names should be tokenized but not stemmed
+      'author_text_nostem_im' => author_primary, # primary author tokenized but not stemmed
+      'sw_author_tesim' => author_primary,
       'contributor_orcids_ssim' => orcids,
       'sw_display_title_tesim' => title,
       'sw_subject_temporal_ssim' => stanford_mods_record.era_facet,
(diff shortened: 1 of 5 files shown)


app/services/author_builder.rb
```diff
@@ -1,34 +1,25 @@
 # frozen_string_literal: true
 
 class AuthorBuilder
-  ALLOWED_ROLES = %w[Author creator].freeze
```
ndushay, line 4: this isn't used anywhere

app/services/author_builder.rb
```diff
@@ -1,34 +1,25 @@
...
 
-  # @param [Array<Cocina::Models::Contributor>] contributors
-  # @return [String] the author value for Solr
-  def self.build(contributors)
```
ndushay, line 8: changing the methos to build_primary and build_all, instead of trying to make "build" serve both cases

app/services/author_builder.rb
```diff
@@ -1,34 +1,25 @@
...
-    contributor = primary_contributor || contributors.first
-    build_contributor(contributor)
+  def build_all
+    NameBuilder.build_all(cocina_contributors.filter_map(&:name))
   end
 
   private
 
-  attr_reader :contributors
+  attr_reader :cocina_contributors
```
ndushay, line 20: let's be clear about what "contributor" means here

app/services/name_builder.rb
```diff
@@ -3,13 +3,21 @@
...
   # @return [Array<String>] names
-  def self.build(names, strategy: :first)
+  def self.build_all(cocina_contributors)
+    flat_names = cocina_contributors.filter_map { |cocina_contributor| flat_names_for(cocina_contributor) }.flatten
+    flat_names.filter_map { |name| build_name(name) }
+  end
+
+  # @param [Symbol] strategy ":first" is the strategy for how to choose a name if primary and display name is not found
+  # @return [String] name
+  def self.build_primary_name(names, strategy: :first)
```
ndushay, line 13: again switching to build_all and build_primary_name

app/indexers/descriptive_metadata_indexer.rb
```diff
@@ -20,7 +20,9 @@ def to_solr
       'mods_typeOfResource_ssim' => resource_type,
       'sw_format_ssim' => sw_format,
       'sw_genre_ssim' => stanford_mods_record.sw_genre,
-      'sw_author_tesim' => author,
+      'contributor_text_nostem_im' => author_all, # author names should be tokenized but not stemmed
+      'author_text_nostem_im' => author_primary, # primary author tokenized but not stemmed
```
ndushay, line 24: if we're keeping sw_author for some reason, let's not stem.  :-)

justinlittman (approved):

PR #1060, opened 2023-11-16
Cache collections

justinlittman:
To reduce the load on DSA with redundant calls.

Unit

### DIFF
--- a/app/services/document_builder.rb
+++ b/app/services/document_builder.rb
@@ -45,25 +45,46 @@ class DocumentBuilder
     Cocina::Models::ObjectType.collection => COLLECTION_INDEXER
   }.freeze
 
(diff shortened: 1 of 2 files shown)


edsu:
So this cache will live as long as the process lives?

justinlittman:
@edsu Yes. That seems safe enough. If it proves problematic, we can come up with a more sophisticated strategy.

edsu:
I guess I just worry about how we will notice that it's become problematic.

edsu (approved):
I'm approving this, but w/ reservations because I'm worried that implementing caching this way could lead to bugs that are difficult to diagnose. I also think that maybe we should think about putting a caching layer in front of DSA. But I realize these aren't practical things to consider while things are on fire.

ndushay (approved):
approving but would like another reviewer as I am also listening to standup

justinlittman:
For what it is worth, we do this sort of caching for APOs in other parts of the code.

edsu:
Do APOs change at roughly the same rate as Collections?

PR #1061, opened 2023-11-16
Re-calibrate the rolling indexer.

justinlittman:
To slow down the rolling indexer is configurable ways.

QA


ndushay:
since parallel may be desired in the future, could this be either 1) deployed as a branch for testing or PR / commit message say "removing parallel for rolling indexer"

justinlittman:
@ndushay I added it to the commit message.

Adding parallel is trivial; there is no reason to leave it in if we don't have immediate uses for it.

ndushay (approved):

PR #1081, opened 2024-01-09
rolling_index: improve performance

ndushay:

See https://docs.google.com/document/d/1B61r-E9v2WhYQ_ABP-RTX61xfbEQ25H8Sm0Wwu5zENA

This improves the performance of the rolling reindexer, and also gives us a few more breadcrumbs in its logs and easier tweaking going forward.

It is strongly recommended the changes in PR sul-dlss/sul-solr-configs/pull/287 are applied to the relevant Solr indexes before deploying this change.  (they have been applied to `qa` and `stage` and `prod` as of this writing.)

In qa, first with batch size of 50 and then batch size of 500.
In stage, with batch size of 500.
In prod, with batch size of 500

In all cases, the new settings have resulted in NOT sending duplicate indexing requests to Solr.  Essentially, soft commits are the bomb!

(compared druids and timestamps in rolling_index log with http://sul-solr-prod-h.stanford.edu/solr/argo_qa/select?q=*:*&facet.range=timestamp&f.timestamp.facet.range.start=NOW%2FDAY-90DAYS&f.timestamp.facet.range.end=NOW&f.timestamp.facet.range.gap=%2B1DAY&rows=501&fl=id,timestamp&facet.field=timestamp&wt=xml&sort=timestamp%20asc)

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,8 +2,12 @@
 date_format_str: '%Y-%m-%d %H:%M:%S.%L'
 rolling_indexer:
   batch_size: 500
-  pause_time_between_docs: .2
-  pause_time_between_batches: 0
+  # in seconds
+  pause_time_between_docs: 0.2
+  # commitWithin does a soft commit by default, so a little more than commitWithin is desired (seconds)
+  pause_time_between_batches: 11
+  # milliseconds
+  commitWithin: 1000
 
 ssl:
   cert_file: ~
--- a/bin/rolling_index
+++ b/bin/rolling_index
@@ -3,7 +3,7 @@
 require_relative '../config/environment'
 require 'daemons'
 
-QUERY = { q: '*:*', sort: 'timestamp asc', fl: 'id', rows: Settings.rolling_indexer.batch_size }
+QUERY = { q: '*:*', sort: 'timestamp asc', fl: 'id,timestamp', rows: Settings.rolling_indexer.batch_size }
 
 Daemons.run_proc(
   File.basename(__FILE__),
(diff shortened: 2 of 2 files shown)


config/settings.yml
```diff
@@ -2,8 +2,12 @@
 date_format_str: '%Y-%m-%d %H:%M:%S.%L'
 rolling_indexer:
   batch_size: 500
-  pause_time_between_docs: .2
-  pause_time_between_batches: 0
+  # in seconds
+  pause_time_between_docs: 0.2
+  # commitWithin does a soft commit by default, so a little more than commitWithin is desired (seconds)
+  pause_time_between_batches: 11
```
justinlittman, line 8: Isn't adding an 11 second pause going to really slow things down?

config/settings.yml
```diff
@@ -2,8 +2,12 @@
 date_format_str: '%Y-%m-%d %H:%M:%S.%L'
 rolling_indexer:
   batch_size: 500
-  pause_time_between_docs: .2
-  pause_time_between_batches: 0
+  # in seconds
+  pause_time_between_docs: 0.2
+  # commitWithin does a soft commit by default, so a little more than commitWithin is desired (seconds)
```
mjgiarlo, line 7: If `commitWithin` is 1s and we want a value that is "a little more," I wonder if it is worth testing other values here to see the effect on 1) duplicates included across batches, and 2) throughput. E.g., the old value of 5s and perhaps 2s. Also happy to ship this PR and have these tweaks be follow-up work if someone is inclined. 🤷🏻‍♂️

ndushay: Yes - and I was experimenting last night, and I will do more today.  (It's in the google doc).  1.2 seconds was too short.  :-)

mjgiarlo: Do we want a value a little more than `commitWithin` or a little more than `autoSoftCommit/maxTime` (from the Solr config)?

config/settings.yml
ndushay: a pause long enough so we don't repeat a 3.7 minute iteration saves us a LOT of time.  I am still experimenting with shorter pauses.   1.2 seconds was too short -- we got some repeats.

ndushay (commented):
I've been keeping notes in the google doc.  

I want to experiment further to see if I can increase throughput, e.g.

- is 500 the right batch size?  (where did that number come from?)
- how small of a pause between batches avoids repeated docs?  (is it really the commitWithin number or the softCommit max time in sul-solr-config)
- is 0.2 seconds between docs a good number?
- should the softCommit max time be lowered from 10 seconds?

It seems I should make an issue in this app for this, as neither of you seems to have read the google doc.

mjgiarlo:
@ndushay 💬 
> It seems I should make an issue in this app for this, as neither of you seems to have read the google doc.

I was in the process of reading the doc, so no need IMO

mjgiarlo (approved):

PR #1095, opened 2024-01-26
use dor_indexing v1.2.0 and cocina-models 0.94.2

ndushay:

To get new title indexing in dor_indexing gem 1.2.0, which uses new methods in cocina-models 0.94.2

CI is enough - we're just indexing a few new title fields.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'sneakers', '~> 2.11'
 gem 'solrizer'
 
 # DLSS gems
-gem 'dor_indexing', '~> 1.1'
+gem 'dor_indexing', '~> 1.1', '>= 1.2.0'
 gem 'dor-services-client', '~> 14.1'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'sneakers', '~> 2.11'
 gem 'solrizer'
 
 # DLSS gems
-gem 'dor_indexing', '~> 1.1'
+gem 'dor_indexing', '~> 1.1', '>= 1.2.0'
```
jmartin-sul, line 23: not a blocker, but i think you could just do `'~> 1.2'`?

jmartin-sul (approved):

Gemfile
jmartin-sul: i made this change just now in a follow-on: https://github.com/sul-dlss/dor_indexing_app/pull/1096

PR #1096, opened 2024-01-26
Gemfile: simplify dor_indexing version specification

jmartin-sul:

i'm pretty sure that what's there can be stated a bit more simply, [explanation from a random stackoverflow denizen](https://stackoverflow.com/a/5170592):

> It means "equal to or greater than in the last digit", so e.g. `~> 2.3` means "equal to 2.3 or greater than 2.3, but less than 3.0", while `~> 2.3.0` would mean "equal to 2.3.0 or greater than 2.3.0, but less than 2.4.0".

`bundle install` worked as expected locally, and updated `Gemfile.lock` constraints without updating the actual version in use (since that just happened in #1095)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'sneakers', '~> 2.11'
 gem 'solrizer'
 
 # DLSS gems
-gem 'dor_indexing', '~> 1.1', '>= 1.2.0'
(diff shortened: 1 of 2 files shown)


ndushay (approved):
cool - i wondered about this.

PR #1102, opened 2024-02-08
remove unused gems from Gemfile (and indirectly loaded gems)

ndushay:

I think some of these gems should have been removed when dor_indexing gem was created.  Others should never have been added.  And others are probably vestigial.

specs and some of the integration tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,8 +6,6 @@ gem 'bootsnap', '>= 1.1.0', require: false # Reduces boot times through caching;
 gem 'bunny', '~> 2.19' # for rabbitmq
 gem 'committee' # Validates HTTP requests/responses per OpenAPI specification
 gem 'config'
-gem 'dry-monads', '~> 1.3'
-gem 'erubis'
 gem 'faraday'
 gem 'honeybadger'
 gem 'okcomputer' # for monitoring
@@ -16,7 +14,6 @@ gem 'rack-timeout', '~> 0.5.1'
 gem 'rails', '~> 7.0.0'
 gem 'rsolr', '~> 2.0'
 gem 'sneakers', '~> 2.11'
-gem 'solrizer'
 
 # DLSS gems
 gem 'dor_indexing', '~> 1.2'
@@ -29,7 +26,6 @@ group :development, :test do
   gem 'rspec_junit_formatter'
   gem 'rspec-rails', '~> 5.0'
   gem 'simplecov'
-  gem 'webmock'
 end
 
 group :development do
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -6,8 +6,6 @@ gem 'bootsnap', '>= 1.1.0', require: false # Reduces boot times through caching;
 gem 'bunny', '~> 2.19' # for rabbitmq
 gem 'committee' # Validates HTTP requests/responses per OpenAPI specification
 gem 'config'
-gem 'dry-monads', '~> 1.3'
```
ndushay, line 9: not sure why dry-monads is explicitly mentioned here;  assume it's leftover from dor_indexing gem extraction

Gemfile
```diff
@@ -6,8 +6,6 @@ gem 'bootsnap', '>= 1.1.0', require: false # Reduces boot times through caching;
...
-gem 'erubis'
```
ndushay, line 10: erubis has to do with erb creation, AFAICT, and isn't needed here.

Gemfile
```diff
@@ -16,7 +14,6 @@ gem 'rack-timeout', '~> 0.5.1'
 gem 'rails', '~> 7.0.0'
 gem 'rsolr', '~> 2.0'
 gem 'sneakers', '~> 2.11'
-gem 'solrizer'
```
ndushay, line 19: again, left here from dor_indexing gem extraction

Gemfile
```diff
@@ -29,7 +26,6 @@ group :development, :test do
   gem 'rspec_junit_formatter'
   gem 'rspec-rails', '~> 5.0'
   gem 'simplecov'
-  gem 'webmock'
```
ndushay, line 32: not used

Gemfile
```diff
@@ -41,9 +37,7 @@ group :development do
 end
 
 group :deployment do
-  gem 'capistrano', '~> 3.0'
   gem 'capistrano-bundler'
   gem 'capistrano-passenger'
-  gem 'capistrano-shared_configs'
```
ndushay, line 47: these two are required by dlss-capistrano, so no need to explicitly require

Gemfile
edsu: Is it not needed for https://github.com/sul-dlss/gis-robot-suite/blob/main/spec/spec_helper.rb#L4

edsu (changes requested):

Gemfile
ndushay: @edsu this is dor-indexing-app;  the spec you mentino is in gis-robot-suite.  I'm confused.

edsu: oops!!!

edsu (approved):

PR #1112, opened 2024-03-28
Use lightweight callables that invoke DSC for indexing

mjgiarlo:
# Why was this change made?

We're currently using `cocina_repository` for both Cocina and non-Cocina concerns, and using a different solution entirely for release tags. This makes each of these independent and switches to an interface that allows lightweight callables to be used over full classes.

# How was this change tested?

CI


app/services/indexer.rb
```diff
@@ -86,12 +92,28 @@ def workflow_client
...
+      raise DorIndexing::RepositoryError, e.message
+    end
+  end
+
+  def release_tags_repository
+    lambda do |druid|
+      Dor::Services::Client.object(druid).release_tags.list
+    rescue Dor::Services::Client::Error => e
+      raise DorIndexing::RepositoryError, e.message
+    end
```
mjgiarlo, line 116: I'll add test coverage back for when these callables fail.

mjgiarlo: @justinlittman I looked into adding test coverage here, but these lambdas are invoked within the dor_indexing gem, not with the app, so to test this in DIA would involve testing the internals of the gem. I could spend time doing that but it seems like something we wouldn't ordinarily do. Advise? If you agree it's beyond the scope of this PR and/or not worth the time, feel free to approve/merge?

justinlittman (approved):

